PR #13, opened 2017-01-04
Single point of dependency management 

dotsbb:
_Please review after #12_ 

Remove dependency management logic out of `travis` and `tox`.
`setup.py` can take care of all needed requirements by itself.

There was an error in `setup.py` since logic was duplicated but not synchronised with `travis` one:
lack of `openid` package in `setup.py` dependencies for `pypy`.

PS: it seems that there is no segfault on travis' python3.5 anymore.


ticosax:
I didn't realized you already did some improvement in this regard.
#15 seems duplicated work at the difference that travis will use tox, which I think is better because we will leverage maintenance of tests execution with tox only.

dotsbb:
I agree that it's better to use `tox` for tests execution. I will close my PRs, but make sure that yours contains fixes for following problems with dependencies:

- different SAML versions for python 2/3/pypy
- test dependencies contains only packages needed to perform tests: it means that  SAML [should be removed](https://github.com/python-social-auth/social-core/pull/13/commits/477792862a3cf94cc791aef215cad992c9223659) from test requirements, as it already  belongs to `extras_require[all,saml]`
- [there was](https://github.com/python-social-auth/social-core/pull/13/commits/e178d2b224cf26b9b6c1f468408e0cc400d31eee) lack of `python-openid` dependency for `pypy`
- install **all** extras in tox: `pip install .[all]`

PR #17, opened 2017-01-15
Including testkey.pem

clintonb:
This file is needed by OpenIdConnectTestMixin. Any packages/projects wanting to use the mixin need the test key.


clintonb:
@omab please review (and release)

omab:
Thanks!

PR #23, opened 2017-01-21
response content of QQ should be decoded first

ZhiweiWang:
I have to decode response content first, as the type of response.content is 'bytes' but the JSON object requires 'str'.


omab:
Thanks!

PR #25, opened 2017-01-22
Add mediawiki backend

tobias47n9e:
If anyone would like to test the backend send me an email. I have a auth-only test application on English Wikipedia and can share the key and secret. It would be good to review the data written to the database and the security and correctness of the pipeline.

### DIFF
--- a/social_core/backends/mediawiki.py
+++ b/social_core/backends/mediawiki.py
@@ -1,0 +1,175 @@
+"""
+MediaWiki OAuth1 backend
+========================
+
+Usage
+-----
+
+In addition to the general setup you need to define the
+following parameters. In Django's settings.py you would
+use for English Wikipedia::
+
+    SOCIAL_AUTH_MEDIAWIKI_KEY = <consumer_key>
+    SOCIAL_AUTH_MEDIAWIKI_SECRET = <consumer_secret>
+    REQUEST_TOKEN_URL =
+        'https://en.wikipedia.org/w/index.php?title=Special:OAuth/initiate'
+    AUTHORIZATION_URL =
+        'https://en.wikipedia.org/w/index.php?title=Special:Oauth/authorize'
+    ACCESS_TOKEN_URL =
+        'https://en.wikipedia.org/w/index.php?title=Special:Oauth/token'
+    IDENTIFY_URL =
+        'https://en.wikipedia.org/w/index.php'
+
+If users should only be prompeted once for permission use:
+
+    AUTHORIZATION_URL =
+        'https://en.wikipedia.org/w/index.php?title=Special:Oauth/authenticate'
+
+General documentation
+---------------------
+
+https://www.mediawiki.org/wiki/Extension:OAuth
+
+Developer documentation
+-----------------------
+
+https://www.mediawiki.org/wiki/OAuth/For_Developers
+"""
+
+import jwt
+import requests
+import six
+from requests_oauthlib import OAuth1
+from six import b
+from six.moves.urllib.parse import parse_qs, urlencode
+from .oauth import BaseOAuth1
+
+
+class MediaWiki(BaseOAuth1):
+
+    """
+    Handles the handshake with Mediawiki and fetching of user data.
+    """
+
+    name = 'mediawiki'
+    REQUEST_TOKEN_URL = 'https://en.wikipedia.org/w/index.php' + \
+                        '?title=Special:OAuth/initiate'
+    AUTHORIZATION_URL = 'https://en.wikipedia.org/w/index.php' + \
+                        '?title=Special:Oauth/authorize'
+    ACCESS_TOKEN_URL = 'https://en.wikipedia.org/w/index.php' + \
+                       '?title=Special:Oauth/token'
+    IDENTIFY_URL = 'https://en.wikipedia.org/w/index.php' + \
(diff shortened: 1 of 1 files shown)


social_core/backends/mediawiki.py
```diff
@@ -0,0 +1,175 @@
...
+    REQUEST_TOKEN_URL =
+        'https://en.wikipedia.org/w/index.php?title=Special:OAuth/initiate'
+    AUTHORIZATION_URL =
+        'https://en.wikipedia.org/w/index.php?title=Special:Oauth/authorize'
+    ACCESS_TOKEN_URL =
+        'https://en.wikipedia.org/w/index.php?title=Special:Oauth/token'
+    IDENTIFY_URL =
+        'https://en.wikipedia.org/w/index.php'
+
+If users should only be prompeted once for permission use:
```
bd808: The mwoauth library took a slightly different approach and chose to always use the `authenticate` endpoint (See https://github.com/mediawiki-utilities/python-mwoauth/commit/52285284254179d2990bb59cac5ef2605b6b9ced). One benefit of this is that much of the work you are doing later to extract the title and merge it into other parameters is eliminated. Instead that library only needs a base URI like `https://en.wikipedia.org/w/index.php` and it can derive all of the URLs needed from that.

social_core/backends/mediawiki.py
```diff
@@ -0,0 +1,175 @@
...
+        oauth_token_key = credentials.get(b('oauth_token'))[0]
+        oauth_token_secret = credentials.get(b('oauth_token_secret'))[0]
+        oauth_token_key = oauth_token_key.decode()
+        oauth_token_secret = oauth_token_secret.decode()
+
+        access_token = {'oauth_token': oauth_token_key,
+                        'oauth_token_secret': oauth_token_secret}
+        return access_token
+
+    def get_user_details(self, token):
```
bd808: The signature of `BaseAuth::get_user_details` is `(self, response)`, not `(self, token)`.

social_core/backends/mediawiki.py
```diff
@@ -0,0 +1,175 @@
...
+                                 auth=auth)
+
+        try:
+            identity = jwt.decode(response.content, secret,
+                                  audience=key, algorithms=['HS256'],
+                                  leeway=10.0)
+        except jwt.InvalidTokenError as exception:
+            raise Exception('An error occurred while trying to read json ' +
+                            'content: {0}'.format(exception))
+
```
bd808: Issuer (`iss`), issue time (`iat`), and `nonce` should be validated from the decoded identity to ensure that the token actually comes from the expected server and is not being replayed somehow. See https://github.com/mediawiki-utilities/python-mwoauth/blob/master/mwoauth/functions.py#L191 for examples.

social_core/backends/mediawiki.py
```diff
@@ -0,0 +1,175 @@
...
+    SOCIAL_AUTH_MEDIAWIKI_KEY = <consumer_key>
+    SOCIAL_AUTH_MEDIAWIKI_SECRET = <consumer_secret>
+    REQUEST_TOKEN_URL =
+        'https://en.wikipedia.org/w/index.php?title=Special:OAuth/initiate'
+    AUTHORIZATION_URL =
+        'https://en.wikipedia.org/w/index.php?title=Special:Oauth/authorize'
+    ACCESS_TOKEN_URL =
+        'https://en.wikipedia.org/w/index.php?title=Special:Oauth/token'
+    IDENTIFY_URL =
+        'https://en.wikipedia.org/w/index.php'
```
tgr: Missing a line here?

social_core/backends/mediawiki.py
tobias47n9e: Good idea to document how the grant system works.

Will only store the base URL in the settings.

social_core/backends/mediawiki.py
tobias47n9e: That makes sense. Will implement the same functions here.

social_core/backends/mediawiki.py
```diff
@@ -0,0 +1,175 @@
...
+    IDENTIFY_URL = 'https://en.wikipedia.org/w/index.php' + \
+                   '?title=Special:OAuth/identify'
+
+    def unauthorized_token(self):
+        """
+        Return request for unauthorized token (first stage)
+
+        Mediawiki request token is requested from e.g.:
+         * https://en.wikipedia.org/w/index.php?title=Special:OAuth/initiate
+         * 'callback_uri' needs to be 'oob'
```
tobias47n9e: I added the option to set the callback in the settings and updated the docs. Thanks for the explanation. I had some trouble understanding that before.

tobias47n9e:
@omab Can you take a look at patch and see if it needs any changes?

omab:
@tobias47n9e thanks, I'll take a look ASAP.

omab:
@tobias47n9e I've reviewed the code and after applying PEP8 and small cleanups I've merged it, also I've moved the docs on the top of the module to [social-docs](https://github.com/python-social-auth/social-docs/blob/master/docs/backends/mediawiki.rst) which renders at https://python-social-auth.readthedocs.io/en/latest/backends/mediawiki.html.

omab:
Thanks!

tobias47n9e:
@omab Thank you too!

Issue #24, opened 2017-01-22
1.0.0 doesn't come with the correct python-openid version (python 3)


khamaileon:
-python3-openid==3.0.10    # via social-auth-core
+python-openid==2.2.5      # via social-auth-core

camilonova:
This problem is for someone getting this:
```
Traceback (most recent call last):
  File "manage.py", line 9, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.5/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.5/site-packages/django/core/management/__init__.py", line 327, in execute
    django.setup()
  File "/usr/local/lib/python3.5/site-packages/django/__init__.py", line 18, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python3.5/site-packages/django/apps/registry.py", line 108, in populate
    app_config.import_models(all_models)
  File "/usr/local/lib/python3.5/site-packages/django/apps/config.py", line 202, in import_models
    self.models_module = import_module(models_module_name)
  File "/usr/local/lib/python3.5/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 986, in _gcd_import
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/usr/local/lib/python3.5/site-packages/social_django/models.py", line 10, in <module>
    from .storage import DjangoUserMixin, DjangoAssociationMixin, \
  File "/usr/local/lib/python3.5/site-packages/social_django/storage.py", line 8, in <module>
    from social_core.storage import UserMixin, AssociationMixin, NonceMixin, \
  File "/usr/local/lib/python3.5/site-packages/social_core/storage.py", line 12, in <module>
    from openid.association import Association as OpenIdAssociation
  File "/usr/local/lib/python3.5/site-packages/openid/__init__.py", line 52, in <module>
    if len(version_info) != 3:
TypeError: object of type 'map' has no len()
```

[...]

PR #26, opened 2017-01-22
Fix the python-openid version

alexandcote:
Remove duplicated requirements
Related to #24.


alexandcote:
I don't follow the contribution guideline, sorry for that...

PR #27, opened 2017-01-22
Fix the python-openid version

alexandcote:
Reordering dependencies
Related to the #24

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,15 +42,14 @@ def read_tests_requirements(filename):
 
 PY = sys.version_info[0]
 requirements_base = read_requirements('requirements-base.txt')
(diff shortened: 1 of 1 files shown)


omab:
This change doesn't fix the issue, the problem wasn't in the order the parameters were read, that shouldn't matter at all and inspecting the files you see that nothing defined in the `base` one is overridden in the python-specific ones.

The problem with the release `1.0.0` was tied with my local environment, the build script (`make build`) just invokes the `setup.py` with `python`, so whatever python version was active at the moment, was then taken by `sys.version_info`, which is directly used to determine the requirements file to load.

I've released `1.0.1` with the fix to this problem.

Thanks

PR #28, opened 2017-01-26
removed IDP dummy settings while generating metadata xml

ziafazal:
Instead of passing dummy IDP settings to `OneLogin_Saml2_Settings` we can use  `sp_validation_only`. That would enable us to set values of `SECURITY_CONFIG` in SP metadata xml.
@omab could you please review this?


omab:
Thanks!

bradenmacdonald:
@ziafazal Nice fix, thanks :)

PR #34, opened 2017-02-01
Added the ability to specify a 'team' in the Slack backend and add in the ID field if it's not present. Fixes #33 #35

### LINKED ISSUE #33: Add the ability to enforce 'team' restriction on Slack auth
Simple addition of a param to pass to the Slack auth URL.

Schwanksta:


Schwanksta:
I updated this pull request to reflect a bug I came across trying to implement Slack auth. I am not sure if Slack sends different JSON formats for different API calls, but the default call didn't return in the UID in a way we could pull it back out.

social_core/backends/slack.py
```diff
@@ -21,6 +21,12 @@ class SlackOAuth2(BaseOAuth2):
         ('real_name', 'real_name')
     ]
 
+    def auth_extra_arguments(self):
+        params = super(SlackOAuth2, self).auth_extra_arguments() or {}
+        if self.setting('team', None):
```
omab: Setting names should be all uppercase.

omab: I've fixed and applied PEP8.

omab:
Thanks!

PR #36, opened 2017-02-01
Allow extra arguments for azuread backend.

jordij:
Calling the `auth_extra_arguments()` method of the parent class allows to add extra params to the request. Handy to include `login_hint` and `domain_hint` among others.

Azure AD oauth call docs: https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-protocols-oauth-code#request-an-authorization-code


omab:
Thanks!

PR #41, opened 2017-02-17
add get_access_token method

NBajanca:
This PR refers to [this Issue](https://github.com/python-social-auth/social-app-django/issues/17).

It is missing tests (@omab I need some help on how to do it if it is even possible).


NBajanca:
At least for Django, I've availability to write a test for this method.

With that done I can add this to the documentation.

omab:
Thanks!

PR #44, opened 2017-02-22
Use "X if Y else Z" over difficult to read quickly "Y and X or Z".

lamby:
(There are some others but it requires X to be "true" so unwilling to
change)

Signed-off-by: Chris Lamb <chris@chris-lamb.co.uk>


omab:
Thanks!

PR #46, opened 2017-02-25
Mediawiki: Raise exception if response is error

tobias47n9e:
If an error occurs during the initial handshake, the error
message is currently not visible to developers.
This commit checks if the response starts with an error
message and raises an exception.


omab:
Thanks!

tobias47n9e:
Thanks for the review & merge!

PR #60, opened 2017-04-13
Note `id_token` in GooglePlusAuth's AuthMissingParameter

### LINKED ISSUE #59: GooglePlusAuth Backend Doesn't Mention `id_token` in Its AuthMissingParameter Exception
Makes it a bit hard to track down errors when the exception doesn't note that using `id_token` is possible.

chrisguitarguy:
So users who do see that error know that `id_token` is supported.

Closes #59


omab:
Thanks!

PR #63, opened 2017-04-14
The Steam backend should use steamid instead of personaname to identify users

stefangeyer:
Use the correct attribute to identify Steam users.

Closes [#64](https://github.com/python-social-auth/social-core/issues/64)

### DIFF
--- a/social_core/backends/steam.py
+++ b/social_core/backends/steam.py
@@ -24,7 +24,7 @@ class SteamOpenId(OpenIdAuth):
         })
         if len(player['response']['players']) > 0:
             player = player['response']['players'][0]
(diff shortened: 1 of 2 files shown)


Matoking:
To elaborate on this, changing your Steam nickname/personaname is as easy as right-clicking your icon in the friend list and selecting "Change Profile Name...". Compared to other online services changing your nickname is very common, which makes the current implementation pretty useless as far as identification is concerned.

As for SteamID, [Valve's own documentation](https://developer.valvesoftware.com/wiki/SteamID) states very clearly that it is an unique identifier used to identify a Steam account.

PR #65, opened 2017-04-16
Speed up authorization process for VKAppOAuth2.

desecho:
The current solution is not working for me. I am always getting empty results here https://github.com/python-social-auth/social-core/blob/master/social_core/backends/vk.py#L136 
and it fails with error because the return of the function is None.


omab:
@desecho, could you add a changelog entry?

PR #67, opened 2017-04-18
add extra_data method to process SAMLAuth

dblado:
Add SAML attributes passed from IDP to extra_data on UserSocialAuth
model. Should behave as other backends do.


omab:
Thanks!

PR #71, opened 2017-04-24
Avoid reusing verification codes

nijel:
Once the code has been used, it should not validate the address again.

Maybe I'm missing something in the code, but it seems that the verified attribute is not really used and because of this the validation codes are valid even after they have been used and it's possible to reuse the partial URLs.


omab:
Thanks!

PR #74, opened 2017-05-02
Updated BaseAuth.authenticate() to prefer kwargs

### LINKED ISSUE #73: Backend authenticate kwargs ignored
The kwargs passed to `BaseAuth.authenticate()` are ignored: https://github.com/python-social-auth/social-core/blob/c03e669a1f98f03e9a838ce3a798a25ad53e9455/social_core/backends/base.py#L73-L74. This is specifically a problem for my team in Django. At this point in the code `self.strategy.request` is `None`. However, the strategy passed in as a kwarg is the instance that has been used for much of the process prior to the call to `authenticate`. `kwargs['strategy'].request` contains the request made to the server to complete authentication.

This is important for my team because we rely on the request for multi-tenancy. We pull the site/tenant from the strategy's request, and from there get the associated configuration: https://github.com/edx/ecommerce/blob/master/ecommerce/social_auth/strategies.py#L12.

I propose modifying

```python
self.strategy = self.strategy or kwargs.get('strategy')
self.redirect_uri = self.redirect_uri or kwargs.get('redirect_uri')
```

to

```python
self.strategy = kwargs.get('strategy', self.strategy)
self.redirect_uri = kwargs.get('redirect_uri', self.redirect_uri)
```

clintonb:
The strategy and redirect_uri passed as kwargs are now preferred to those values set in the constructor. The primary goal of this change is to ensure that the strategy has a request attached so that downstream code can make use of the request.

Fixes #73

### DIFF
--- a/social_core/backends/base.py
+++ b/social_core/backends/base.py
@@ -70,8 +70,8 @@ class BaseAuth(object):
            'strategy' not in kwargs or 'response' not in kwargs:
             return None
 
-        self.strategy = self.strategy or kwargs.get('strategy')
-        self.redirect_uri = self.redirect_uri or kwargs.get('redirect_uri')
+        self.strategy = kwargs.get('strategy') or self.strategy
+        self.redirect_uri = kwargs.get('redirect_uri') or self.redirect_uri
         self.data = self.strategy.request_data()
         kwargs.setdefault('is_new', False)
         pipeline = self.strategy.get_pipeline(self)


clintonb:
@omab please review/merge

social_core/backends/base.py
```diff
@@ -70,8 +70,8 @@ def authenticate(self, *args, **kwargs):
            'strategy' not in kwargs or 'response' not in kwargs:
             return None
 
-        self.strategy = self.strategy or kwargs.get('strategy')
-        self.redirect_uri = self.redirect_uri or kwargs.get('redirect_uri')
+        self.strategy = kwargs.get('strategy', self.strategy)
+        self.redirect_uri = kwargs.get('redirect_uri', self.redirect_uri)
```
omab: Can you change it to be:

```python
self.strategy = kwargs.get('strategy') or self.strategy
self.redirect_uri = kwargs.get('redirect_uri') or self.redirect_uri
```

clintonb: Done.

omab:
Thanks!

clintonb:
@omab thanks for merging. I'm always happy to contribute. Any idea on when you'll publish a new release to PyPI?

omab:
I'll park some time to do it this week. Thanks for the contributions, they are always welcome.

PR #75, opened 2017-05-02
Implemented ORCID backend

andreyyudin:
Implemented ORCID OAuth 2 backend (https://orcid.org/) and its sandbox option. Separately doing a pull request for the documentation in social-docs.


omab:
@andreyyudin could you add a changelog entry?

PR #79, opened 2017-05-09
Fix auth header formatting problem. (fitbit)

edelvalle:
Before the results was like:

`{'Authorization': 'Basic b"o9834yf0398efyq30br87fyb30f8u3hyb"'}`

There was the `b""` Python formating for byte strings. So I added a decode step.

Final result is like:

`{'Authorization': 'Basic o9834yf0398efyq30br87fyb30f8u3hyb'}`


omab:
@edelvalle could you add a changelog entry?

edelvalle:
@omab done!

PR #81, opened 2017-05-14
Raise AuthForbidden on 401.

mhozza:
This causes middlewere to catch the exception,
which prevents the server error, for example when
user presses back button on browser and repeaats the request
with already used grant code.


omab:
@mhozza could you add an entry to the changelog?

omab:
Thanks!

PR #82, opened 2017-05-15
Add udata/datagouvfr backend

davidbgk:
Add a generic base class for [udata](http://udata.readthedocs.io/en/latest/) backend, set as a base class given that French are not the only ones using that platform.

I need some help to understand how tests are working. I don't get how to set up the username for instance, do I have to create a dummy account on the platform?

(Incoming social-docs PR counter-part)


omab:
@davidbgk, usually what I do is log the response from the usual flow, then copy and paste the response into the test class and overwrite any sensitive information.

davidbgk:
@omab tests are now passing!

omab:
@davidbgk, thanks!

PR #84, opened 2017-05-21
Maintain `attributes` key when sending in response for SAML

UmanShahzad:
Without this, one cannot eventually retrieve `attributes` by referencing the key because it's already dereferenced. It's hard to get what you want through the [base `extra_data` method](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/base.py#L113) when your [`EXTRA_DATA`](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/saml.py#L172) list has to specify all the specific received keys, instead of just specifying "`attributes`".

Either this, or make the internal `EXTRA_DATA` have "attributes" as an element and make the method just pass in `response` without any dereferencing, like so:

```python
EXTRA_DATA = ['attributes']

...

def extra_data(self, user, uid, response, details=None, *args, **kwargs):
    return super(SAMLAuth, self).extra_data(user, uid, response, details=details,
                                            *args, **kwargs)
```

### DIFF
--- a/social_core/backends/saml.py
+++ b/social_core/backends/saml.py
@@ -312,7 +312,7 @@ class SAMLAuth(BaseAuth):
 
     def extra_data(self, user, uid, response, details=None, *args, **kwargs):
         return super(SAMLAuth, self).extra_data(user, uid,
-                                                response['attributes'],
+                                                {'attributes': response['attributes']},
                                                 details=details,
                                                 *args, **kwargs)
 


haikuginger:
@UmanShahzad, I understand what you're trying to do here, but I think the change introduced in #67 is actually the correct behavior. What would actually be useful to our use case would be to make an optional "wildcard" value available for `EXTRA_DATA` which, if present, results in all keys being sent back. Such an implementation would also be generic across the board, rather than a very specific hack for SAML.

@omab, would a PR implementing such a wildcard be something you'd be open to? Perhaps we could implement a change right [here](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/base.py#L120) that checks for a singleton `ALL_FIELDS` object which, if present, causes `data.update(response)` (adding each value with its name as sent on the wire to the data dictionary).

UmanShahzad:
@haikuginger That's a great alternative. I would definitely want to implement that if @omab agrees to the proposal.

omab:
I like the proposal for a configurable flag to fetch all the entries in the response and store them in extra data. I've left a comment on #86, but the proposal is in the right path. This one can be closed, right?

UmanShahzad:
@omab It's good to close :+1:

PR #88, opened 2017-06-01
Fix DisqusOAuth2 backend since the callback URL is now validated

mback2k:
Disqus now returns the following error if redirect_state is used:

Invalid Request: Invalid parameter: redirect_uri (should match predefined callback URI)


omab:
Thanks!

PR #89, opened 2017-06-07
fix weixin backend request parameter

cheng10:
Weixin oauth doc requires an additional paramter:  'scope': 'snsapi_login'. 
Otherwise, it would not work.

### DIFF
--- a/social_core/backends/qq.py
+++ b/social_core/backends/qq.py
@@ -59,7 +59,7 @@ class QQOAuth2(BaseOAuth2):
         response = self.get_json(
             'https://graph.qq.com/user/get_user_info', params={
                 'access_token': access_token,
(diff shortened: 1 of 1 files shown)


social_core/backends/weixin.py
```diff
@@ -53,7 +53,8 @@ def auth_params(self, state=None):
         appid, secret = self.get_key_and_secret()
         params = {
             'appid': appid,
-            'redirect_uri': self.get_redirect_uri(state)
+            'redirect_uri': self.get_redirect_uri(state),
+            'scope': 'snsapi_login',
```
omab: This should be in the `DEFAULT_SCOPE` attribute in the class, the lib will pick the value and add it to the URL automatically.

PR #90, opened 2017-06-07
Adds Patreon OAuth2 backend

xfxf:
As per documentation at https://www.patreon.com/platform/documentation/oauth

We were using this with an older python-social-auth (0.2.21), but I've ported this forward to social-core so upstream benefits and it's not stuck in our forked version.

Tests should pass - am having issues running tox locally, but will see what your CI (assuming Travis) does.


xfxf:
Tests appear to now pass, so hopefully good to merge!

omab:
Thanks!

PR #92, opened 2017-06-08
Sanitize disconnect URLs as well

nijel:
Apply same sanitization as on connect to disconnect.


nijel:
Originally reported against Weblate: https://hackerone.com/reports/238117

omab:
Thanks!

PR #98, opened 2017-06-21
Update withings URLs, now they belong to Nokia.

edelvalle:


omab:
@edelvalle I wasn't aware the was a PR already for this when I committed the change a few days ago. Thanks anyway!

PR #102, opened 2017-06-23
Update get_attr to use the entire value if an attribute is single-valued

### LINKED ISSUE #101: SAML2 NameID only picks first letter
When using the SAML2 backend, the get_attr function will return just the first letter of the nameid.
This is caused by NameID always being a single-valued field, whereas other attributes are always provided as a list, because they can have multiple values.
This means that the `attributes[key][0]` in get_attr returns the first letter.

puiterwijk:
NameID is a single-valued value, which means that when [0] is applied to it, we grab
just the first letter instead of the full nameid.

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

Fixes #101


omab:
@puiterwijk could you add an entry to the changelog?

puiterwijk:
@omab done so

omab:
Thanks!

PR #106, opened 2017-06-28
Feature/azuread tenant

sdementen:
add support for tenants in Azure AD backend


omab:
I've made some changes to the backend and proceeded to merge it:

* Split into `azuread_tenant.py` module
* Extra name to install needed dependencies (cryptography)

PR #118, opened 2017-07-24
fix weixin default scope

cheng10:
Weixin oauth api doc requires an additional paramter: 'scope': 'snsapi_login'.
Otherwise, it would not work.


social_core/backends/weixin.py
```diff
@@ -17,6 +17,7 @@ class WeixinOAuth2(BaseOAuth2):
     AUTHORIZATION_URL = 'https://open.weixin.qq.com/connect/qrconnect'
     ACCESS_TOKEN_URL = 'https://api.weixin.qq.com/sns/oauth2/access_token'
     ACCESS_TOKEN_METHOD = 'POST'
+    DEFAULT_SCOPE = 'snsapi_login'
```
omab: `DEFAULT_SCOPE` should be a tuple or array, otherwise the code will fail by trying to concatenate an string into an array.

omab:
I've made the requested change and proceeded to merge this. Thanks!

Issue #128, opened 2017-09-20
UserMixin.expiration_timedelta() returns wrong value because timedelta is calculated using different timezones


DmitriySelischev:
I am using google oauth2 with django. The UserMixin provides expiration check by access_token_expired() that executes expiration_timedelta() inside.
So, the problem is in expiration_timedelta. Code of the method:
```python
        if self.extra_data and 'expires' in self.extra_data:
            try:
                expires = int(self.extra_data.get('expires'))
            except (ValueError, TypeError):
                return None

            now = datetime.utcnow()

            # Detect if expires is a timestamp
            if expires > time.mktime(now.timetuple()):
                # expires is a datetime, return the remaining difference
                return datetime.fromtimestamp(expires) - now
            else:
                # expires is the time to live seconds since creation,
                # check against auth_time if present, otherwise return
                # the value
                auth_time = self.extra_data.get('auth_time')
                if auth_time:
                    reference = datetime.fromtimestamp(auth_time)
                    return (reference + timedelta(seconds=expires)) - now
                else:
                    return timedelta(seconds=expires)

```

The problem occurs because we use UTC now time for current moment, but we use LOCAL time when executing `datetime.fromtimestamp`.
How it works?
Imagine that I am in +2 timezone and right now I have a 12 o'clock UTC and expiration time is 3600 seconds. I got access_token and refresh_token at 11 o'clock UTC, so, my token is already expired. But, when I get auth_time then I'll get 13 o'clock LOCAL + expiration time = 14 o'clock. Timedelta is 14 o'clock LOCAL subtract 12 o'clock UTC = 2 hours(It's strange, but python calculates it like these datetimes at the same timezone). And then we check expiration_timedelta().total_seconds() <= 0 and we got false instead of true.

The fix is very simple: use `datetime.utcfromtimestamp` instead of `datetime.fromtimestamp`.
I do that expiration check manually in my project right now and all works smoothly.

I'll provide a pull request with fix.

omab:
Fixed by #129

PR #129, opened 2017-09-22
Fixed issue #128 with expiration_timedelta calculation.

DmitriySelischev:
Issue #128 fixed. Replaced fromtimestamp to utcfromtimestamp


omab:
Merged, thanks!

PR #132, opened 2017-09-23
Add fields to /me request in user_data in Facebook backend

### LINKED ISSUE #131: Facebook backend not retrieving email
We recently experienced an issue where the response from the Facebook graph API at `/me` just after authenticating didn't contain an email field. It's hard to say if this issue is present for everyone, or whether it's just an issue with our site - in Ruby's `omniauth`, users experiencing the issue seemed to be experiencing it intermittently: https://github.com/mkdynamic/omniauth-facebook/issues/61

Adding `SOCIAL_AUTH_FACEBOOK_PROFILE_EXTRA_PARAMS = {'fields': 'id,name,email,first_name,last_name'}` to our settings made it work.

To avoid this problem in the future, perhaps it would be good to have the `user_data` method make the request with a reasonable default set of fields, including email?

donaldharvey:
Attempt to fix #131.

### DIFF
--- a/social_core/backends/facebook.py
+++ b/social_core/backends/facebook.py
@@ -64,7 +64,8 @@ class FacebookOAuth2(BaseOAuth2):
 
     def user_data(self, access_token, *args, **kwargs):
         """Loads user data from service"""
-        params = self.setting('PROFILE_EXTRA_PARAMS', {})
+        params = {'fields': 'id,name,email,first_name,last_name'}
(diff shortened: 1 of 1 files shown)


omab:
The values in `fields` to be passed to `/me` are directly related to the scope specified, for instance, in order to get the user email you need the email scope set too, that's reason to make it open for developers to define the value instead of trying to set a value that attempt to make everybody happy.

Thanks!

Issue #133, opened 2017-09-26
Use constant time compare in state validation 


peterholczbauer-prezi:
Due to [security reasons](https://cryptocoding.net/index.php/Coding_rules#Compare_secret_strings_in_constant_time) `constant_time_compare` should be used to compare the two `state` parameters within the `validate_state` method instead of the `==` in the `OAuthAuth ` class ([here](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/oauth.py#L90)).

`constant_time_compare` is available in the [`social-core/utils.py`](https://github.com/python-social-auth/social-core/blob/master/social_core/utils.py#L213).

```
elif not state:
    raise AuthStateMissing(self, 'state')
elif not constant_time_compare(request_state,state):
    raise AuthStateForbidden(self)
else:
    return state
```

PR #134, opened 2017-09-26
Fix Issue #133: Constant time compare for state validation

peterholczbauer-prezi:
Issue #133 fixed:
Use the `constant_time_compare` method from the `social_core.utils` package to prevent [timing attacks](https://cryptocoding.net/index.php/Coding_rules#Compare_secret_strings_in_constant_time) against the state token during the state token validation.


omab:
Merged, thanks!

PR #143, opened 2017-10-19
reddit wants authorization at token stage

maparent:
I had an unauthorized for https://ssl.reddit.com/api/v1/access_token that was solved by this change.


omab:
Thanks!

PR #144, opened 2017-10-19
Use extras_require for version differences

### LINKED ISSUE #142: social_core setup.py may detect wrong python version when using pipenv
With the new pipenv, which is [recommended by python.org now](https://packaging.python.org/new-tutorials/installing-and-using-packages/#installing-pipenv), there are issues with using os.sys_version in setup.py to determine which dependencies to install.

It appears that since pipenv technically runs in your main python environment rather than inside the virtualenv it manages, setup.py detects that python version.  So for example I use python 3.6 as my main python, but have a few legacy projects which are on 2.7 for now.  If I use pipenv then social_core's setup.py sees python 3.6 in os.sys_version and pops python3-openid into the requirements rather than python-openid.

Would it be possible to use the `extras_require` feature of setuptools 18.0 and up to handle the conditional dependency?

Please also see https://github.com/kennethreitz/pipenv/issues/924

jmichalicek:
Updated setup.py to use extras_require to differentiate between python 2 and python 3 requirements.  This gets along with pipenv. Fixes #142 

Now pipenv users may add some additional specifications to their Pipfile to get the correct versions installed consistently.

```
[packages]
social-auth-core = "*"
python-openid = {version=">=2.2.5", markers="python_version<'3.0'"}
python3-openid = {version=">=3.0.10", markers="python_version>='3.0'"}
defusedxml = {version=">=0.5.0rc1", markers="python_version>='3.0'"}
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,9 +44,11 @@ def read_tests_requirements(filename):
 
 
 PY = os.environ.get("BUILD_VERSION") or sys.version_info[0]
-requirements_base = read_requirements('requirements-base.txt')
-requirements = requirements_base + \
-    read_requirements('requirements-python%s.txt' % PY)
+requirements = read_requirements('requirements-base.txt')
+# May be able to just use environment markers in requirements-base.txt
+# at least on  setuptools 36.2.0 and up.
+requirements_py2 = read_requirements('requirements-python2.txt')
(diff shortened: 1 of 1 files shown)


omab:
Thanks!

PR #147, opened 2017-10-28
Update Facebook backend REDIRECT_STATE to False

### LINKED ISSUE #141: social_core.backends.facebook.FacebookOAuth2 should have REDIRECT_STATE set to False
I'm new to Facebook Oauth and actually new to the entire Oauth world, so this bug took me very long time to figure out.

Since I'm new I'm forced to use Facebook API v2.10.

When `REDIRECT_STATE` set to `True`, PSA will compose authentication url like following:

`https://www.facebook.com/v2.10/dialog/oauth?scope=email&redirect_uri=http%3A%2F%2Fexample.com%2Fcomplete%2Ffacebook%2F%3Fredirect_state%3D1TIlS0FPkTctrPCQhgf2rdO3bawWvwdB&client_id=[client_id]&return_scopes=true&state=1TIlS0FPkTctrPCQhgf2rdO3bawWvwdB
`

And Facebook will reject such kind of url, gives error message like following:

> Can't Load URL: The domain of this URL isn't included in the app's domains. To be able to load this URL, add all domains and subdomains of your app to the App Domains field in your app settings.

This the callback url configured on redirect URL doesn't allow `redirect_state` inside the URL, renders the authentication url invalid.

I'm not sure if older version of Facebook API allows this, but this is the case I'm encountering, and when I override the setting `REDIRECT_STATE` with a custom backend, Facebook allows me in.

Following is the entire code I'm using now with Facebook API v2.10:

```python
from social_core.backends.facebook import FacebookOAuth2

class FacebookOAuth2Override(FacebookOAuth2):
    REDIRECT_STATE = False
```

And settings:
```python
SOCIAL_AUTH_FACEBOOK_KEY = 'key'
SOCIAL_AUTH_FACEBOOK_SECRET = 'secret'
SOCIAL_AUTH_FACEBOOK_SCOPE = ['email', 'public_profile']
SOCIAL_AUTH_FACEBOOK_PROFILE_EXTRA_PARAMS = {
  'locale': 'en',
  'fields': 'id,name'
}
```

Please verify and fix, cheers.

piotr-szpetkowski:
Fix #141

### DIFF
--- a/social_core/backends/facebook.py
+++ b/social_core/backends/facebook.py
@@ -20,6 +20,7 @@ API_VERSION = 2.9
 class FacebookOAuth2(BaseOAuth2):
     """Facebook OAuth2 authentication backend"""
     name = 'facebook'
+    REDIRECT_STATE = False
     RESPONSE_TYPE = None
     SCOPE_SEPARATOR = ','
     AUTHORIZATION_URL = 'https://www.facebook.com/v{version}/dialog/oauth'


KaczuH:
When it will be merged?
I need it.

tzangms:
Hi, Facebook just announced that strict mode will be forced in 90 days, hope this `REDIRECT_STATE=False` be merged soon, or everyone will need to customize a FB Backend to accommodate this.

tisdall:
A note should probably added to the CHANGELOG.md .  Essentially this is a needed change for ALL Facebook oAuth since without this change all requests will fail after they enforce a strict redirect URI matching.

omab:
Version `1.6.0` published into pypi.

Issue #145, opened 2017-10-19
Partial decorator should be parameterised: do not save the token into the session


peterholczbauer-prezi:
**Issue**
Saving the `partial token` into the session is not always a good thing to do.
[Done here](https://github.com/python-social-auth/social-core/blob/master/social_core/pipeline/partial.py#L35)

**Why**
For example: Google provides the feature to add other gmail accounts to your "default" account. When a user tries to authenticate with one of the accounts, but the pipeline is interrupted with a partial step, he/she cannot authenticate with another account, as long as the "first" `partial_token` is in the session.

**Propose**
 The decorator is supposed to be parametrised to be able to control whether saving the token into the session or not.

**Solution**
```
def partial(save_to_session=True):
    def decorator(func):
        @wraps(func)
        def wrapper(strategy, backend, pipeline_index, *args, **kwargs):
            current_partial = partial_prepare(strategy, backend, pipeline_index,
                                              *args, **kwargs)

            out = func(strategy=strategy,
                       backend=backend,
                       pipeline_index=pipeline_index,
                       current_partial=current_partial,
                       *args, **kwargs) or {}

            if not isinstance(out, dict):
                strategy.storage.partial.store(current_partial)
                if save_to_session:
                    strategy.session_set('partial_pipeline_token', current_partial.token)
            return out
        return wrapper
    return decorator
```
 
**WDYT?**
If you agree I can submit a PR with the "fix".

omab:
Yes, I agree, the saving in the session is there to be backward compatible, making that an option sounds good to me.

PR #149, opened 2017-11-02
Fix issue: #145 - Make the partial step decorator handle arguments

peterholczbauer-prezi:
Create a new `pratial_step` decomaker, that makes the `partial` decorator handle arguments.
With the new decorator saving the `partial token` into the session is optional.
Keep the `partial` decorator backward compatible.

Issue Fixed: #145


omab:
Thanks!

PR #153, opened 2017-11-08
Fix coinbase backend to use api v2

joshblum:
Current implementation seems to be broken. https://developers.coinbase.com/api/v2#show-a-user


omab:
Thanks!

PR #155, opened 2017-11-14
WIP: Add JIRA backend

nkprince007:


nkprince007:
@omab, some help here, please?

requirements-base.txt
```diff
@@ -3,3 +3,4 @@ oauthlib>=1.0.3
 requests-oauthlib>=0.6.1
 six>=1.10.0
 PyJWT>=1.4.0
+cryptography>=2.1.1
```
omab: Where's this used?

requirements-python3.txt
```diff
@@ -1,3 +1,4 @@
 defusedxml>=0.5.0rc1
 python3-openid>=3.0.10
+pycryptodome>=3.4.7
```
omab: Where's this used?

social_core/backends/jira.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def get_unauthorized_token(self):
+        return self.get_json(
+            self.REQUEST_TOKEN_URL,
+            auth=self.oauth_auth()
+        )
+
+    @property
+    def ACCESS_TOKEN_URL(self):
+        return self._ACCESS_TOKEN_URL.format(**self.jira_url_params)
```
omab: There's no need to define this property, you can define the method `access_token_url` instead.

social_core/backends/jira.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @property
+    def AUTHORIZATION_URL(self):
+        return self._AUTHORIZATION_URL.format(**self.jira_url_params)
```
omab: There's no need to define this property, you can define the method `authorization_url` instead.

requirements-python3.txt
nkprince007: It's being used for RSA1 encryption.

nkprince007: It is an external dependency which oauthlib isn't bundled with.

requirements-base.txt
nkprince007: Please check https://github.com/python-social-auth/social-core/pull/155#discussion_r157049337

nkprince007:
@omab please check the replies on review comments. I've updated the pull request with the changes you've suggested.

Now for handling JIRA Server REST APIs, each request sent to the server should be signed by an RSA priv/pub key pair which has already been registered as an incoming Application link in JIRA administration.

So, these additional requirements, `pycryptodome` and `cryptography` are required for RSA 1 signing. These algorithms are not bundled along with `oauthlib` and need to be installed separately.

PR #160, opened 2017-11-28
Fix AOL OpenID auth

osya:
Changes for the AOL OpenID auth mostly based on the corresponding LiveJournal OpenID auth code. Three tests written (according to the LiveJournalOpenIdTest). Two of them `test_login` and `test_failed_login` passed. Test `test_partial_pipeline` failed. IMHO it is due to wrong response message from AOL (which contains in the `server_response` variable in tests). For example, OpenID namespace 2.0 should be contained in this message or `openid.claimed_id` should be changed to to the '`openid1_claimed_id`

### DIFF
--- a/social_core/backends/aol.py
+++ b/social_core/backends/aol.py
@@ -2,9 +2,24 @@
 AOL OpenId backend, docs at:
     https://python-social-auth.readthedocs.io/en/latest/backends/aol.html
 """
+from urllib.parse import urlsplit
+
+from social_core.exceptions import AuthMissingParameter
+
 from .open_id import OpenIdAuth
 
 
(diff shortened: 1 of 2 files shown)


omab:
@osya what's the reasoning for this change? Why drop generic openid auth with users based ones that's not backward compatible?

osya:
@omab, the reason for this change is just existing code for AOL OpenID in the module `aol.py` is not worked. I began investigate it and in the end write this PR. With my changes AOL OpenID worked somehow (two of three tests written by me according to the tests for LiveJournalOpenIdTest - passed).

nijel:
The service is gone.

PR #169, opened 2017-12-26
Add Phabricator backend

divadsn:
This adds a OAuth2 backend for Phabricator to allow users to authenticate using Phabricator login.

See: https://phabricator.wikimedia.org/T179202

### DIFF
--- a/social_core/backends/phabricator.py
+++ b/social_core/backends/phabricator.py
@@ -1,0 +1,46 @@
+"""
+Phabricator OAuth2 backend, docs at:
+    https://secure.phabricator.com/book/phabcontrib/article/using_oauthserver/
+"""
+from .oauth import BaseOAuth2
+
+
+class PhabricatorOAuth2(BaseOAuth2):
+    """Phabricator OAuth authentication backend"""
+    name = 'phabricator'
+    API_URL = 'https://secure.phabricator.com'
+    AUTHORIZATION_URL = 'https://secure.phabricator.com/oauthserver/auth/'
+    ACCESS_TOKEN_URL = 'https://secure.phabricator.com/oauthserver/token/'
+    ACCESS_TOKEN_METHOD = 'POST'
+    REDIRECT_STATE = False
+
+    def api_url(self, path):
+        api_url = self.setting('API_URL') or self.API_URL
+        return '{0}{1}'.format(api_url.rstrip('/'), path)
+
+    def authorization_url(self):
+        return self.api_url('/oauthserver/auth/')
+
+    def access_token_url(self):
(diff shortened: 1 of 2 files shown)


social_core/backends/phabricator.py
```diff
@@ -0,0 +1,48 @@
...
+
+class PhabricatorOAuth2(BaseOAuth2):
+    """Phabricator OAuth authentication backend"""
+    name = 'phabricator'
+    API_URL = 'https://secure.phabricator.com'
+    AUTHORIZATION_URL = 'https://secure.phabricator.com/oauthserver/auth/'
+    ACCESS_TOKEN_URL = 'https://secure.phabricator.com/oauthserver/token/'
+    ACCESS_TOKEN_METHOD = 'POST'
+    REDIRECT_STATE = False
+    DEFAULT_SCOPE = []
```
jayvdb: If there is no default scope, this could also be omitted, or better set to `None` same as the super class, but ... 

I suspect that the default scope is actually `scope.always`.  Check out  https://secure.phabricator.com/D15621 , which removed previous scopes `offline_access` and `whoami` as no longer needed, but added a new one.

Phabricator has no published defined scopes, as noted at https://github.com/ofbeaton/oauth2-phabricator#managing-scopes , and "This section has not been written yet." on https://secure.phabricator.com/book/phabcontrib/article/using_oauthserver/ .

If true, this deserves special mention as Phabricator is very odd in that regard; not many other OAuth providers have no scopes at all.

But as I mentioned above, it looks like the scope `scope.always` exists, and I wouldnt be surprised if the myriad other [scopes in phabricator](https://github.com/phacility/phabricator/search?utf8=%E2%9C%93&q=scope&type=) also work via OAuth.

social_core/backends/phabricator.py
```diff
@@ -0,0 +1,48 @@
...
+            response.get('realName')
+        )
+
+        return {
+            'id': response.get('phid'),
+            'username': response.get('userName'),
+            'email': response.get('primaryEmail', ''),
+            'fullname': fullname,
+            'first_name': first_name,
+            'last_name': last_name
```
jayvdb: trailing commas are a good idea ; look at where else you might add them

divadsn:
I squashed those styling edits back into two commits @jayvdb :)

jayvdb:
It is one feature; only one commit is necessary

divadsn:
Hm, I was taking #162 as an example for my commits, but I can squash them later into one, I still have not figured out how to use git on a Chromebook :D

omab:
Thanks!

PR #176, opened 2018-01-16
updated code to fallback on userPrincipalName for email - Microsoft backend

jpatel3:
'`mail`' field is often null when the accounts are not native Microsoft one, and in those cases, `userPrincipalName` field is used. You can find some more information on [here](https://stackoverflow.com/questions/47344384/microsoft-azure-ad-graph-api-how-do-i-retrieve-a-users-email-address). With this change it tries to get `email` from `userPrincipalname` if the `mail` field is empty. (The check for the type of `userPrincipalName` can be improved, although if `userPrincipalName` is not email, in that case, you can expect that value of `mail` will be there. )

### DIFF
--- a/social_core/backends/microsoft.py
+++ b/social_core/backends/microsoft.py
@@ -46,9 +46,13 @@ class MicrosoftOAuth2(BaseOAuth2):
     def get_user_details(self, response):
         """Return user details from Microsoft online account"""
         email = response.get('mail')
+        username = response.get('userPrincipalName')
         if not email and '@' in response.get('userPrincipalName'):
             email = response.get('userPrincipalName')
-        return {'username': response.get('displayName', ''),
+        if '@' in username:
+            username = username.split('@')[0]
+
+        return {'username': username,
                 'email': email,
                 'fullname': response.get('displayName', ''),
                 'first_name': response.get('givenName', ''),
--- a/social_core/backends/microsoft.py
+++ b/social_core/backends/microsoft.py
@@ -45,8 +45,11 @@ class MicrosoftOAuth2(BaseOAuth2):
 
     def get_user_details(self, response):
         """Return user details from Microsoft online account"""
+        email = response.get('mail')
+        if not email and '@' in response.get('userPrincipalName'):
+            email = response.get('userPrincipalName')
         return {'username': response.get('displayName', ''),
-                'email': response.get('mail'),
+                'email': email,
                 'fullname': response.get('displayName', ''),
                 'first_name': response.get('givenName', ''),
                 'last_name': response.get('surname', '')}


jpatel3:
@Menda no problem. How does the release cycle work? (when it will be available)

Menda:
I don't know. It's up to the maintainer to merge or not.

social_core/backends/microsoft.py
```diff
@@ -45,8 +45,11 @@ def get_user_id(self, details, response):
 
     def get_user_details(self, response):
         """Return user details from Microsoft online account"""
+        email = response.get('mail')
+        if not email and '@' in response.get('userPrincipalName'):
+            email = response.get('userPrincipalName')
         return {'username': response.get('displayName', ''),
-                'email': response.get('mail'),
+                'email': email,
```
jefmoura: * Based on the Microsoft documentation email and logon namespaces are basically the same. So, I think you can always use the UPN instead.

```
"The point of the UPN is to consolidate the email and logon namespaces so that the user only needs
to remember a single name."
```
[**User Naming Attributes** - userPrincipalName (Microsoft)](https://msdn.microsoft.com/en-us/library/ms677605(v=vs.85).aspx#userPrincipalName)

social_core/backends/microsoft.py
```diff
@@ -45,8 +45,11 @@ def get_user_id(self, details, response):
 
     def get_user_details(self, response):
         """Return user details from Microsoft online account"""
+        email = response.get('mail')
+        if not email and '@' in response.get('userPrincipalName'):
+            email = response.get('userPrincipalName')
         return {'username': response.get('displayName', ''),
```
jefmoura: * `displayName` is the user's name, e.g., John Lennon. Shouldn't the username property be set to a normal one like `johnlennon` instead? If yes, we can just use the UPN because it's also the username for Microsoft.

social_core/backends/microsoft.py
jpatel3: If I understand correctly, UPN won't always email. In that case, it may cause the wrong entries. From their documentation, it was hard to find the accurate answer as in different app context it was written differently.

social_core/backends/microsoft.py
omab: Looking at the docs, that looks exactly what's needed here, the `UPN` should be used as the `username`, if it contins an `@` on it, then the prefix needs to be extracted and used instead.

jpatel3:
@omab Updated as requested, please review it.

omab:
Merged, thanks!

jpatel3:
Thanks @omab  When it's going to be incorporated in the release?

jpatel3:
@omab Any plan for the release in coming months?

Issue #178, opened 2018-01-17
LinkedIn OAuth2 broken


makslevental:
repeatedly get

`requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: https://api.linkedin.com/v1/people/~:(email-address,first-name,id,last-name)?format=json&oauth2_access_token=<snip>`

problem is here:

https://github.com/python-social-auth/social-core/blob/master/social_core/backends/linkedin.py#L88

it should be

```python
        return self.get_json(
            self.user_details_url(),
            params={
              'format': 'json',
              'oauth2_access_token': access_token
            }
        )
```

JVUnderground:
Can we update the pip package, so we don't have to use unofficial installs?

gennaroanesi:
getting this same issue here. got it working with @makslevental correction 👌

makslevental:
any movement on this? this is a pretty big deal isn't it?

Josue23:
Getting this same issue here.
Where do I put this code from @makslevental?

makslevental:
@Josue23 https://github.com/python-social-auth/social-core/pull/179/files

gennaroanesi:
any news on this one?

drewjaja:
Currently using social core 1.5 to get around the issue.

omab:
I've just released version `1.7.0` which carries the `Bearer` authentication method, that's the correct format documented by LinkedIn ([Step 4 — Make authenticated requests](https://developer.linkedin.com/docs/oauth2)). `oauth2_access_token` header is not needed.

PR #179, opened 2018-01-17
Update linkedin.py

makslevental:
fix https://github.com/python-social-auth/social-core/issues/178


yprez:
LGTM :+1: 

@omab would be great to get this merged :)

omab:
This change is not needed at all:
1. Sending the access token as `Bearer` header is the right format (and works properly), check [Step 4 — Make authenticated requests](https://developer.linkedin.com/docs/oauth2)
2. There's a `facebook` backend change unrelated and also not needed, collisions on email are triggered by the pipeline and shouldn't be handled on a per-backend basis.

PR #182, opened 2018-01-24
Added LinkedIn mobile authentication backend

axper:
LinkedIn uses 2 separate access tokens for their mobile sdk client and the web client.

Currently python-social-auth uses the web client. This means that in a project which has user account registrations possible from both a web view and a mobile client, when the user registers in the mobile client, he will not be able to make authentication requests to the web server, since the mobile access key doesn't work for web requests.

This merge request solves the issue by adding a separate authentication backend which imitates LinkedIn mobile SDK requests.

https://stackoverflow.com/a/39032743/2529583
https://stackoverflow.com/a/41221636/2529583

Please tell me if there are any improvements required.


omab:
Thanks!

PR #183, opened 2018-01-24
Add aliased extra_data field on refresh_token

uda:
When updating extra_data in `UserMixin.refresh_token`, some fields with data get overridden with `None`, because they are stored with the alias and not the original name (like `expires_in` → `expires` and `id` → `user_id` for google plus).

So this should fix those cases.


omab:
Thanks!

Issue #185, opened 2018-01-24
OAuth2 with jive fails with 411 because parameters are passed as data even in GET case


maparent:
I am connecting to a Jive account. I left the default ACCESS_TOKEN_METHOD=GET.
The BaseOAuth2.auth_complete method passes params as data, not params, so the GET has a body; but requests (2.10.0, I know....) seems not to send Content-Length because it's a GET. Sending a PR so the params are sent as params in GET case.

PR #186, opened 2018-01-24
Use params instead of data in token request, if the method is GET.

maparent:
I had a case of 411 (Content-Length required) on Jive because the parameters were sent as data. #185


omab:
Thanks!

PR #189, opened 2018-02-06
azuread_b2c policy name check fix

robertavram:
Fix, policy name could start with capitalized letters "B2C_"


OmarIthawi:
I'm thinking that maybe we should add a test case for this specific method. It would be great if you can do that @robertavram.

omab:
Thanks!

PR #193, opened 2018-02-12
Allow for cases where SAML user_permanent_id is not a list

maparent:
In that case, the method as originally given returns the first character of the UID string instead of the full UID.


omab:
Thanks!

PR #200, opened 2018-02-21
Remove additional 'redirect_state' parameter from GitHub backend

fermayo:
GitHub understands the `state` parameter on their OAuth flows:
https://developer.github.com/apps/building-oauth-apps/authorization-options-for-oauth-apps/#web-application-flow

The current configuration using `redirect_state` actually breaks OAuth flows using GitHub Apps as they [don't support any extra parameters in the redirect URI](https://developer.github.com/apps/building-github-apps/identifying-and-authorizing-users-for-github-apps/)


omab:
Thanks!

PR #204, opened 2018-02-24
Update access token url + authorize url

sanderbrauwers:
https://developers.coinbase.com/docs/wallet/coinbase-connect/integrating


omab:
Thanks!

PR #208, opened 2018-02-26
Fixed literal warnings in slugify

rsalmaso:
Running my test with enabled warnings I have
```bash
../lib/social_core/utils.py:138: DeprecationWarning: invalid escape sequence \w
  value = re.sub('[^\w\s-]', '', value).strip().lower()
../lib/social_core/utils.py:139: DeprecationWarning: invalid escape sequence \s
  return re.sub('[-\s]+', '-', value)
```
This PR uses raw strings.


omab:
Thanks!

PR #211, opened 2018-03-06
social-core#62: sanitizing urls like `///evil.com`

jamesmgg:
we got some malicious redirect on our app. this need to be fixed ASAP.


omab:
Merged, thanks!

PR #212, opened 2018-03-07
Add support for Zendesk oAuth

danielsamuels:


avelis:
@danielsamuels I would like to say thank you for submitting this effort. I had to create something similar.

For anyone else who comes here. One of the main issues is with persisting the subdomain value to make both URL calls to Zendesk (auth + token call). There is also the implicit grant flow that would skip token URL POST call and receive an OAuth token direct to the callback URL. However, the second approach doesn't seem to play well with the current `BaseOauth2` sequence of method calls and would require more fiddling than I am willing to make at the moment. I am not sure if we will ever see a Zendesk provider in the library but I hope one day it will be there.

PR #214, opened 2018-03-15
Add Keycloak backend with tests

aleksihakli:
This patchset adds a Keycloak OAuth 2.0 backend based on the Keycloak OpenID Connect 1.0 API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Delete partial token from session if still present
 - Use `userPrincipalName` to set `username` and `email` accordingly
 - Send authorization headers to Kakao OAuth2, properly fill user details
+- Keycloak backend is added with preliminary OAuth2 support
 
 ## [1.7.0](https://github.com/python-social-auth/social-core/releases/tag/1.7.0) - 2018-02-20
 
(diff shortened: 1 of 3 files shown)


aleksihakli:
It seems like this patch set fails builds on Python 3.3 because Cryptography is not installed in the Tox build environment. [Python 3.3 is EOL](https://www.python.org/download/releases/3.3.0/) and could be deprecated.

aleksihakli:
Hey @omab! Could this be merged into upstream?

aleehedl:
Works well. Hoping to see this merged as soon as possible.

lukaszlacinski:
Keycloak is an OIDC IdP. Why does the backend subclass BaseOAuth2 instead of OpenIdConnectAuth?

aleksihakli:
We needed a authentication of users, and in our case OAuth 2.0 offered that with some customizations. Due to that I went with a KISS approach, just subclassing the OAuth 2.0 authentication, which produced less code and complexity, and appears to work well in the specific use case. OIDC 1.0 is, after all, based on the OAuth 2.0 protocol.

It is possible to implement OIDC 1.0 support as well, if there is a need and an ask for that.

omab:
The change looks great, I might use it as an example in the future when properly documenting how to contribute backends.

Sorry about the delay on merging it.

aleksihakli:
Great to have it upstream :) I'll see about contributing the OIDC backend as well, would be neat to have the full implementation available.

PR #225, opened 2018-04-19
Feature/expanded backend

andreyyudin:
The existing ORCID backend has been expanded to provide an option to use Member API.


omab:
Thanks!

PR #245, opened 2018-06-01
fix qq bug when get access token

haipersist:
when getting access token for qq,it'll raise exception if the data is None.
so we should take a judgement before sending post request.


omab:
Closing, the code is not even properly indented, I don't think this was even tested.

PR #258, opened 2018-07-11
Fix error when using id_token with Google+

Sazpaimon:
The data from id_token contains the scopes the client requested and cannot be used with the Google+ user info API, so instead of hitting that API, just use the data returned when validating the token.


Sazpaimon:
I updated the PR to move the `tokeninfo` call to `user_data` so users implementing the workflow at https://python-social-auth-docs.readthedocs.io/en/latest/use_cases.html#signup-by-oauth-access-token can use `id_token`

omab:
Merged, thanks!

omab:
Strange, the work is merged (https://github.com/python-social-auth/social-core/blob/ef7e104ee763c46bfdf2232ee2537cda4762f8c8/social_core/backends/google.py#L137-L152), but the PR still reports as not merged, I'll just close the PR.

PR #274, opened 2018-08-29
Added support for ELIXIR AAI

tauceti2:


tauceti2:
@omab, please can you have a look on failing tests? Our OIDC server will not allow you to do the /authorize via cmdline, because we are using federated authN. Also for we are not by default allow clients to request any scope, scopes are enabled to the clients manually after the approval. Is it possible to pass the test with this configuration?

tauceti2:
Added missing entry into the change log.

bgruening:
@omab can you please have a look at this PR. Thanks.

bgruening:
@omab thanks a lot!

PR #283, opened 2018-10-17
Update qq.py

talebook:
kwargs already has 'params'.  Current old code cause exception.

      File "/usr/local/lib/python2.7/dist-packages/social_core/backends/qq.py", line 70, in request_access_token
        response = self.request(url, params=data, *args, **kwargs)
    TypeError: request() got multiple values for keyword argument 'params


omab:
Thanks!

PR #285, opened 2018-10-24
More generically detect cancelling in oauth2 process_error

tuky:
This e.g. fixes the flow with linkedin: https://developer.linkedin.com/docs/oauth2#
They either raise `user_cancelled_login` or `user_cancelled_authorize` which are both caught by this change. I would argue, that I did not overgeneralize, but if that's an issue, the error code matching could be narrowed down further.


omab:
Thanks!

Issue #55, opened 2017-04-03
Line oauth should not identify 'statusMessage' as error when getting user data


iamgeorgelee:
https://github.com/python-social-auth/social-core/blob/master/social_core/backends/line.py#L39

I actually get a success response when calling `user_data` like below, but being processed as error because it contains `statusMessage`:

> {u'displayName': u'George Lee',
> u'mid': u'XXXXX',
> u'pictureUrl': XXXXX',
> u'statusMessage': u'My status'}`

iamgeorgelee:
I'll create a PR for the fix later

beoreoji:
Why is it closed? 
This has not been solved yet.

PR #291, opened 2018-11-09
Fix Line process_error bug

beoreoji:


omab:
Any doc pointing to why this change is needed? Any related ticket?

laymonage:
@omab see #55.
`'statusMessage'` is user's profile status, not error status message. I've tested it myself.

PR #292, opened 2018-11-12
MediaWiki: get_user_details should return more details

urbanecm:
get_user_details should return more things from MediaWiki instance:

* email
* confirmed_email
* editcount
* rights
* groups
* registered (time when user registered)
* blocked


omab:
Thanks!

Issue #70, opened 2017-04-24
Uid not set on Azure AD authentication and NoneType error


twschiller:
I'm trying to setup social-core, social-core-django, and Azure AD backend to authenticate using Azure Active Directory. The initial authentication works and creates a User and User Social Auth. However, subsequent attempts to authorize result in:

- A new User object (but no corresponding User Social Auth object)
- The server error 'NoneType' object has no attribute 'provider' raised at social_core/actions.py in do_complete, line 69

Some relevant information:
- The Uid field in the User Social Auth object created from the first Authorization is `None`
- The Azure AD is set up to include the "Sign in and read user profile" delegated permission (which corresponds to the `User.Read` permission)
- [Reading the backend code, it looks like the email address and ID is supposed to come from a field 'upn' corresponding to the user's on-premises user principal name (UPN)](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/azuread.py#L63). The User Name field as listed in Azure AD
- The id_token JWT contains the following fields: aud, iss, iat, nbf, exp, aio, amr, email, family_name, given_name, idp, ipaddr, name, oid, platf, sub, tid, unique_name, uti, and ver. Note that the JWT does not contain the upn.
- The extra data stored with the User Social Auth includes auth_time, access_token, id_token, refresh_token, expires, expires_on, not_before (null), first_name, last_name, token_type, and resource (null)

What I've tried so far:
- I've tried using both the default pipeline, and the pipeline described in the documentation at https://python-social-auth-docs.readthedocs.io/en/latest/configuration/django.html#personalized-configuration 

Key questions:
- How is the Azure AD server supposed to pass the `upn` back to the client? Is it in the JWT?
- Is the missing UPN also the root cause for the NoneType server error on subsequent authentications?

I'm going to follow up with Microsoft Azure AD support to understand why the UPN is not being returned given my configuration, and will include their response

twschiller:
Responses below from MS. Looks like I'm going to have to create a modified version of the azure ad backend that uses oid for uid and the email field for the email

"Given my configuration, should the Azure AD server be returning a UPN?"

> It depends.  If you are authN to AAD as a user of that tenant then you should get the upn claim.  However, if the user is not a member of the tenant or if you’re authN as an application using a key then you won’t.

If so, where would the UPN be returned? Would it be returned in the JWT?

> It would be in the jwt.  https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-token-and-claims

[...]

PR #298, opened 2018-12-16
Update azuread_tenant.py to fix NoneType error

kowenhouston:
Override the get_user_id function from AzureADOAuth2 as the uid does not exist in the azure ad tenant response and produces a "None" UID within the social authentication.  See issue #70 . The Azureb2c backend has the exact same override function. Have tested locally and the changes work as expected.


omab:
Thanks!

Issue #300, opened 2018-12-21
Google+ Starts Going Away in January 2019


pizzapanther:
I think this URL needs to change: https://github.com/python-social-auth/social-core/blob/master/social_core/backends/google.py#L56

From Google:

*On March 7, 2019, all Google+ APIs and Google+ Sign-in will be shut down completely. This will be a progressive shutdown beginning in late January, with calls to these APIs starting to intermittently fail as early as January 28, 2019.*

pizzapanther:
I think this is the URL that needs to be used instead: `https://www.googleapis.com/oauth2/v2/userinfo`

pizzapanther:
or `https://www.googleapis.com/userinfo/v2/me`

codingjoe:
Same here, we just got the warning from google. I would also recommend to remove the Google+ backends in the next release. This way users get yet another warning, that they will be deprecated soon. Very soon in fact.

marc-gist:
thanks @pizzapanther hopefully this makes it into the pip repository asap :)

pizzapanther:
I have a PR coming with just the URL change. Testing now. I figured the other Google+ removal should come later since you can still use it if you really want to until March.

tisdall:
Someone just submitted #301...  maybe you should collaborate, @pizzapanther

omab:
PR #301 merged.

omab:
Also, `v3.0.0` released with the changes.

Tiagojdferreira:
I just started looking at the issue of deprecation of the api and I have some questions regarding the migration process.
I see that the GooglePlusAuth class extends the BaseGoogleOAuth2API (fixing the url), but I also see it still asks for [the scopes `plus.me` and `plus.login` which should be replaced by `profile` and `openid`.](https://developers.google.com/identity/sign-in/web/quick-migration-guide).

Are you completely dropping support for the class GooglePlusAuth and I should start using GoogleOAuth2?

tisdall:
@Tiagojdferreira - There's 2 different classes, `GoogleOAuth2` and `GooglePlusAuth`.  The #301 PR fixes the `GoogleOAuth2` one so it no longer has any connection to Google+ and will continue to work after the shutdown.  The `GooglePlusAuth` continues to use the Google+ API's and will eventually stop working when Google shuts it down.

[...]

PR #301, opened 2018-12-21
Move away from Google+ API usage in google-oauth2 backend

suligap:
Fix for #300.

Google+ APIs will be shut down on March 7, 2019. More details here
https://developers.google.com/+/api-shutdown

> This will be a progressive shutdown, with intermittent failures
> starting as early as January 28, 2019.

Replacing the G+ endpoint with
https://www.googleapis.com/oauth2/v3/userinfo works for the
google-oauth2 backend with almost no extra changes.

This new endpoint is currently used by
[Google's OpenID Connect](https://developers.google.com/identity/protocols/OpenIDConnect)
implementation as OpenID Connect `userinfo_endpoint`. It might not be
the best idea to hard code it like that in the long term. But this was
the one endpoint that is actually mentioned in some Google
documentation as far as I could find. Please feel free to propose better
alternatives if you can find them.

`GooglePlusAuth` backend is affected by this change too.
However according to https://developers.google.com/+/web/signin/
the Google+ Sign-in feature is also fully deprecated and will be shut
down on March 7, 2019. I'd expect for the changes related
dropping/deprecating `GooglePlusAuth` backend in this library to be done
in a separate change. This change makes sure `google-oauth2` backend
will continue to work during and after Google+ APIs deprecation.

After this change this note from the docs can be removed too I believe:

> the application requires Google+ API to be enabled in the Google
> console dashboard

At least as far as the `google-oauth2` backend goes.

### DIFF
--- a/social_core/backends/google.py
+++ b/social_core/backends/google.py
@@ -12,7 +12,10 @@ class BaseGoogleAuth(object):
     def get_user_id(self, details, response):
         """Use google email as unique id"""
         if self.setting('USE_UNIQUE_USER_ID', False):
-            return response['id']
+            if 'sub' in response:
+                return response['sub']
+            else:
+                return response['id']
         else:
             return details['email']
 
@@ -20,24 +23,14 @@ class BaseGoogleAuth(object):
         """Return user details from Google API account"""
         if 'email' in response:
             email = response['email']
-        elif 'emails' in response:
-            email = response['emails'][0]['value']
         else:
             email = ''
 
-        if isinstance(response.get('name'), dict):
-            names = response.get('name') or {}
-            name, given_name, family_name = (
-                response.get('displayName', ''),
-                names.get('givenName', ''),
-                names.get('familyName', '')
-            )
-        else:
-            name, given_name, family_name = (
-                response.get('name', ''),
-                response.get('given_name', ''),
-                response.get('family_name', '')
-            )
+        name, given_name, family_name = (
+            response.get('name', ''),
+            response.get('given_name', ''),
+            response.get('family_name', ''),
+        )
 
         fullname, first_name, last_name = self.get_user_names(
             name, given_name, family_name
@@ -53,7 +46,7 @@ class BaseGoogleOAuth2API(BaseGoogleAuth):
     def user_data(self, access_token, *args, **kwargs):
         """Return user data from Google API"""
         return self.get_json(
-            'https://www.googleapis.com/plus/v1/people/me',
+            'https://www.googleapis.com/oauth2/v3/userinfo',
             params={
                 'access_token': access_token,
                 'alt': 'json'
--- a/social_core/tests/backends/test_google.py
+++ b/social_core/tests/backends/test_google.py
@@ -16,64 +16,23 @@ from .open_id_connect import OpenIdConnectTestMixin, NO_JWKEST
 
 class GoogleOAuth2Test(OAuth2Test):
     backend_path = 'social_core.backends.google.GoogleOAuth2'
-    user_data_url = 'https://www.googleapis.com/plus/v1/people/me'
+    user_data_url = 'https://www.googleapis.com/oauth2/v3/userinfo'
     expected_username = 'foo'
     access_token_body = json.dumps({
         'access_token': 'foobar',
         'token_type': 'bearer'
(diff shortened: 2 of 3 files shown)


social_core/backends/google.py
```diff
@@ -12,32 +12,25 @@ class BaseGoogleAuth(object):
     def get_user_id(self, details, response):
         """Use google email as unique id"""
         if self.setting('USE_UNIQUE_USER_ID', False):
-            return response['id']
+            if 'sub' in response:
```
codingjoe: *EAFP*; I would suggest to use try catch, eg:
```python
try:
    return response['sub']
except KeyError:
    return response['id']
```

social_core/tests/backends/test_google.py
```diff
@@ -16,64 +16,23 @@
...
-        'emails': [{
-            'type': 'account',
-            'value': 'foo@bar.com'
-        }],
-        'etag': '"e-tag string"',
-        'gender': 'male',
-        'id': '101010101010101010101',
-        'image': {
-            'url': 'https://lh5.googleusercontent.com/-ui-GqpNh5Ms/'
+        'profile': 'https://plus.google.com/101010101010101010101',
```
codingjoe: Where is the profile coming from? I don't get it when I do a GET on the `userinfo` endpoint. I get:
```
{
 "id": "**************",
 "email": "*****@johanneshoppe.com",
 "verified_email": true,
 "name": "Johannes Hoppe",
 "given_name": "Johannes",
 "family_name": "Hoppe",
 "link": "https://plus.google.com/+JohannesHoppe1",
 "picture": "https://lh4.googleusercontent.com/-DkWjGAz2yP8/AAAAAAAAAAI/AAAAAAAAEv4/ro4cfX5UqDs/photo.jpg",
 "gender": "male",
 "locale": "en",
 "hd": "johanneshoppe.com"
}
```

social_core/backends/google.py
suligap: I agree. I'd normally do that. But I followed the convention used in this module.

social_core/tests/backends/test_google.py
suligap: Your example response looks like a one coming from a different endpoint than the one used in this PR. https://www.googleapis.com/oauth2/v3/userinfo does return a `profile` field (and has `sub` instead of `id` for example). If you're using the same endpoint, then we're getting different results..

suligap:
Thank you for your reviews and testing @pizzapanther and @codingjoe.

social_core/tests/backends/test_google.py
codingjoe: Sorry, I used v2. All good:
```JSON
{
  "profile": "https://plus.google.com/+JohannesHoppe1", 
  "family_name": "Hoppe", 
  "sub": "***", 
  "picture": "https://lh4.googleusercontent.com/-DkWjGAz2yP8/AAAAAAAAAAI/AAAAAAAAEv4/ro4cfX5UqDs/photo.jpg", 
  "locale": "en", 
  "email_verified": true, 
  "given_name": "Johannes", 
  "gender": "male", 
  "email": "***@johanneshoppe.com", 
  "hd": "johanneshoppe.com", 
  "name": "Johannes Hoppe"
}
```

Pomax:
Is this good to land? =)

nprica3pg:
Is there a release plan for this pull request? It would be nice to have it usable in the near future, since 28th of January is coming soon. Thank you.

tisdall:
Don't wait for this to be merged and released...  Create a file in your own project with these changes and replace the lines of code using `social_core.backends.google.GoogleOAuth2` with your own copy.  I did it by just pulling a copy from https://raw.githubusercontent.com/suligap/social-core/migrate-away-from-google-plus-api/social_core/backends/google.py and changing the imports to:

```
from social_core.utils import handle_http_errors
from social_core.backends.open_id import OpenIdAuth
from social_core.backends.oauth import BaseOAuth2, BaseOAuth1
from social_core.exceptions import AuthMissingParameter
```

When this is merged and released, I can upgrade to the latest, point back at `social_core.backends.google.GoogleOAuth2` and delete my local copy.

omab:
PR merged, thanks for the update.

PR #302, opened 2018-12-22
Pass access token via Authorization header to google user data url

suligap:
i.e. the "default" OAuth2 method of presenting the access token to the
resource server. Which is of course supported by both:

- https://www.googleapis.com/plus/v1/people/me (current user data url)
- https://www.googleapis.com/oauth2/v3/userinfo (see #301)

Passing secrets via GET query params is not great. Resource server
access logs usually retain the entire url including the GET query
params. `requests` library when used with DEBUG logging level will also
log the entire request url including the query params on the OAuth
client side.

This also drops the unnecessary `alt=json` query param.

This change should be compatible with #301.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Updated Azure B2C to extract first email from list if it's a list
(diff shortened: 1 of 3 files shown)


omab:
Merged. Thanks!

PR #303, opened 2019-01-07
Sending profile fetch calls to OAuth2 API instead of Google+ API

shnaqawi:
As Google+ API is getting depreciated. The call for fetching user info could be sent to: https://www.googleapis.com/userinfo/v2/me 
instead of:
https://www.googleapis.com/plus/v1/people/me


tombasche:
This should already be fix as part of https://github.com/python-social-auth/social-core/pull/301

omab:
#301 took care of this and it also provided an update on tests cases. Thanks!

PR #311, opened 2019-01-25
allow to override JWT decode options in OIC backend

skarzi:
Sometimes, we don't want to verify some claims, which verification is often optional (like `sub`).
In current version it's impossible to parametrize `jwt.decode()` and that's the reason I am propsing to add attribute with all options specific to current `OpenIdConnectAuth` subclass, that will be simply passed to `jwt.decode()`.

Thank you for amazing project!


skarzi:
@omab could you please look at this PR?

PR #309, opened 2019-01-22
WIP: Added Auth0 provider

alesrosina:
Hi all, this is my first submission - I wanted to add  provider for Auth0, which I will need in an existing app called Weblate (http://weblate.org). I think it makes sense to make a contribution here, since it might be useful also for others.

Code is actually from: https://auth0.com/docs/quickstart/webapp/django/01-login

PS: This is my first contact with Django framework, so I'm not really 100% confident if I did all correct.

PR #317, opened 2019-02-18
Added function for extra arguments on Auth0 provider

alesrosina:
@omab added a missing function, which got slipped of #309 pull request.

### DIFF
--- a/social_core/backends/auth0.py
+++ b/social_core/backends/auth0.py
@@ -21,6 +21,10 @@ class Auth0OAuth2(BaseOAuth2):
         return 'https://{domain}/{path}'.format(domain=self.setting('DOMAIN'),
                                                 path=path)
 
+    def auth_extra_arguments(self):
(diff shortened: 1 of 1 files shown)


social_core/backends/auth0.py
```diff
@@ -21,6 +21,10 @@ def api_path(self, path=''):
         return 'https://{domain}/{path}'.format(domain=self.setting('DOMAIN'),
                                                 path=path)
 
+    def auth_extra_arguments(self):
+        """Return extra arguments needed on request-token process"""
+        return self.setting('AUTH_EXTRA_ARGUMENTS', {})
```
omab: Why is this function needed? The default defined in the base class will do the same.

alesrosina: Ouch, you are right. I didn't see that in base class ... sorry for that!

alesrosina:
Closing this, since it is a duplication of code.

PR #320, opened 2019-02-19
Update google_openidconnect to use non-deprecated APIs

aleksanb:
In the last PR that updated GoogleOAuth2 to use the non-deprecated APIs
instead of the Google+ ones, google_openidconnect was left on the old
deprecated APIs. I've updated the config now, and tested that this
works locally on our project.

### DIFF
--- a/social_core/backends/google_openidconnect.py
+++ b/social_core/backends/google_openidconnect.py
@@ -12,10 +12,3 @@ class GoogleOpenIdConnect(GoogleOAuth2, OpenIdConnectAuth):
     # differs from value in discovery document
     # http://openid.net/specs/openid-connect-core-1_0.html#rfc.section.15.6.2
     ID_TOKEN_ISSUER = 'accounts.google.com'
-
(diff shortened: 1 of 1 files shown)


social_core/backends/google_openidconnect.py
```diff
@@ -12,10 +12,3 @@ class GoogleOpenIdConnect(GoogleOAuth2, OpenIdConnectAuth):
     # differs from value in discovery document
     # http://openid.net/specs/openid-connect-core-1_0.html#rfc.section.15.6.2
     ID_TOKEN_ISSUER = 'accounts.google.com'
-
-    def user_data(self, access_token, *args, **kwargs):
```
awachat: @aleksanb I am curious to know how to obtain userdata if this code is removed? Someone just made another PR to use new APIs to get userdata. [PR Link](https://github.com/python-social-auth/social-core/pull/330)

omab:
Fixed by #330

Issue #310, opened 2019-01-23
LinkedIn v1 deprecation


mpranjic:
https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/migration-faq?context=linkedin/consumer/context?trk=eml_mktg_gco_dev_api_comms

It seems linkedin backend is using v1 ([link](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/linkedin.py#L15))

danielquinn:
The API will be [dropped](https://engineering.linkedin.com/blog/2018/12/developer-program-updates) as of March 1st (2weeks from now).

I'm looking into this for our own system, and if something useful comes out of it, I'll contribute it back, but it'd be good to know if anyone else is working on this.

W3Technics:
I started implementing this, but I might need some help with the tests. They are kind of confusing me. Can I just open a pull request and let it be reviewed?

W3Technics:
Getting a 'AuthFailed: Authentication failed: ('Connection aborted.', BadStatusLine("''",))' error on the tests, after getting the access_token, kind of lost. If any one could take a look and figure out what I'm doing wrong (probably something silly), that would be great!

jaapdeheer:
@W3Technics I'm also trying to get this to work (as quickly as possible, with dirty hacks if need be), so if you get it working I'll be very interested! Of course I'll shout here if I have something working, though it probably won't be very fit for public consumption :)
Good luck!

danielquinn:
[Here's what I have so far](https://gist.github.com/danielquinn/e5da6dc263a9e691a865932b16b64f3b).  It may be a decent starting point for someone trying to make something more generic.

jaapdeheer:
@danielquinn you're a lifesaver! This seems to work perfectly for me.
Looks like I'll have time for sleep this weekend after all 😁 
Thanks a million to both @danielquinn & @W3Technics for sharing your hard work!

jaapdeheer:
BTW I crudely removed the type hints to make it work in Python 3.4, my modified version is [here](https://gist.github.com/jaapdeheer/b58ee0d6ded142702f88c945c7dd3f78) in case it helps anyone.

W3Technics:
I updated my code ([here](https://github.com/W3Technics/social-core/blob/feature/linkedin_api_v2/social_core/backends/linkedin_v2.py)) the code @jaapdeheer provided, and added support for retrieving profile pictures. I'm still testing this implementation.

[...]

PR #322, opened 2019-02-25
Updated LinkedinOAuth2 to use v2 APIs #310

awachat:
Recently, Linkedin announced that all the developers need to migration to Linkedin v2 API and OAuth2 by March 1st, 2019. 

This PR
- Added support for Linkedin OAuth2.0 using v2 API
- Removed unsupported Oauth1 from the Linkedin backend

These changes do not require any modification to the existing application built with python-social-core.

PS: Also updated the LinkedIn documentation on social-docs. [PR Link](https://github.com/python-social-auth/social-docs/pull/55)

### DIFF
--- a/social_core/backends/linkedin.py
+++ b/social_core/backends/linkedin.py
@@ -20,9 +20,8 @@ class LinkedinOAuth2(BaseOAuth2):
     EXTRA_DATA = [
         ('id', 'id'),
         ('expires_in', 'expires'),
-        ('firstName', 'first_name', True),
-        ('lastName', 'last_name', True),
-        ('emailAddress', 'email')
+        ('firstName', 'first_name'),
+        ('lastName', 'last_name')
     ]
 
     def user_details_url(self):
--- a/social_core/backends/linkedin.py
+++ b/social_core/backends/linkedin.py
@@ -17,6 +17,7 @@ class LinkedinOAuth2(BaseOAuth2):
                       '?q=members&projection=(elements*(handle~))'
     ACCESS_TOKEN_METHOD = 'POST'
     REDIRECT_STATE = False
+    DEFAULT_SCOPE = ['r_liteprofile']
     EXTRA_DATA = [
         ('id', 'id'),
         ('expires_in', 'expires'),
--- a/social_core/backends/linkedin.py
+++ b/social_core/backends/linkedin.py
@@ -2,23 +2,109 @@
 LinkedIn OAuth1 and OAuth2 backend, docs at:
     https://python-social-auth.readthedocs.io/en/latest/backends/linkedin.html
 """
-from .oauth import BaseOAuth1, BaseOAuth2
+from social_core.exceptions import AuthCanceled
+from .oauth import BaseOAuth2
+
+API_VERSION = 2
 
 
-class BaseLinkedinAuth(object):
-    EXTRA_DATA = [('id', 'id'),
-                  ('expires_in', 'expires'),
-                  ('first-name', 'first_name', True),
-                  ('last-name', 'last_name', True),
-                  ('firstName', 'first_name', True),
-                  ('lastName', 'last_name', True)]
-    USER_DETAILS = 'https://api.linkedin.com/v1/people/~:({0})'
+class LinkedinOAuth2(BaseOAuth2):
+    name = 'linkedin-oauth2'
+    AUTHORIZATION_URL = \
+        'https://www.linkedin.com/oauth/v{version}/authorization'
+    ACCESS_TOKEN_URL = 'https://www.linkedin.com/oauth/v{version}/accessToken'
+    USER_DETAILS_URL = \
+        'https://api.linkedin.com/v{version}/me?projection=({projection})'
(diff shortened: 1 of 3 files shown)


danielquinn:
One suggestion I wanna make here is that I think you can reduce a considerable amount of complexity by just dropping the `version` argument.  If the intention is to replace `linkedin-oauth2` with a version that supports the new API, then the only thing you have to worry is about is conserving the contract between this library and those that use it.  So long as the input & output is the same, whichever URL(s) you're talking to aren't relevant, especially since v1 will go offline days if not hours after this is merged into master and is pushed to PyPI.

awachat:
My thought process is to make it easy to integrate with the new versions of LinkedIn API, e.g. v3, v4 by changing just the version number. FacebookOAuth implementation follows a similar versioning design. 

Given a second thought, Facebook changes the API frequently and supports multiple versions, and there are no drastic changes to the consequent API versions. Linkedin is more about maintaining a single version for a long time, and the updates are mostly backward incompatible. 

Given the above context, removing generic versioning from Linkedin backend to simply the code sounds good. Thanks @DanielJDufour

W3Technics:
Great! Thanks, @awachat and @danielquinn!

awachat:
@W3Technics, Thanks for the suggestion. 
I resisted adding `r_emailaddress` in the default scope because of the following reasons.

1.  `LinkedinOauth2` is backward compatible, meaning to use API v2, there is no need to make any changes in the existing code. In the earlier implementation, the default values were `id`, `firstName` and `lastName` & `r_basicprofile` for field_selectors and scope respectively. 
2. Respecting the LinkedIn design to treat email id differently. Email is not mandatory to implement login with `LinkedinOauth2`. There is a possibility that some developers might have used `LinkedinOauth2` without the `r_emailaddress` scope (almost without any customization to social-core). If we add it by default now, it may cause compliance issues in such applications. We have two choices, optimize for most of the developers who will use emails or make it backward compatible with least privileges to not cause any compliance issue unknowingly. I focused on the latter.

danielquinn:
I'm with you on :two:, though it may be a safe bet that nearly everyone will have/want the `r_emailaddress` permission.  I just wanted to point out that everyone will have to adjust their scope values, since `r_basicprofile` has been removed in favour of `r_liteprofile`.

social_core/backends/linkedin.py
```diff
@@ -2,94 +2,108 @@
...
-        if scope:
-            scope = '?scope=' + self.SCOPE_SEPARATOR.join(scope)
-        return self.request(self.REQUEST_TOKEN_URL + scope,
-                            params=self.request_token_extra_arguments(),
-                            auth=self.oauth_auth()).text
-
+        if 'emailAddress' in set(self.setting('FIELD_SELECTORS', [])):
+            emails = self.email_data(access_token, *args, **kwargs)
+            if emails:
+                response['emailAddress'] = emails[0]
```
thomas-scrace: I hit an exception here. `TypeError: 'filter' object is not subscriptable`

Should it be `response['emailAddress'] = next(emails)]` ?

W3Technics: I changed the 'return Filter(None, emailAddresses)' line in the 'email_data' function to just 'return emailAdresses' to work around this. @awachat, what is the use of the filter without a filterfunction?

W3Technics:
@awachat : alright, valid points!

social_core/backends/linkedin.py
thomas-scrace: I think it's just to filter out `None`s.

Maybe call `list()` on the result before returning?

thomas-scrace:
I'm now getting back URN values like `urn:li:digitalmediaAsset:C5603AQFje8e_VeNZUw` for profile image URLs with the new API. Should we consider doing something to convert these to URLs, which is probably what most people are expecting?

(https://developer.linkedin.com/docs/ref/v2/media-migration)

social_core/backends/linkedin.py
awachat: It works in python2, looks like in python3 .filter() does not return a list but an iterator. 

`response['emailAddress'] = next(emails)` -> works only in python3 not in python2. 

Thanks for flagging this.

@W3Technics, `None` filter on a list removes all the falsify values from the list.
```
filter(None, [1, None, "", {}, []])

# return
[1]
```

If there is anything else breaking with python3, please do let me know.

awachat:
@thomas-scrace Checkout [this](https://gist.github.com/awachat/751bf6a354ea414eb48f85af9780b270) gist 
returns firstName, lastName, profilePicture, emailAddress of a LinkedIn user.

omab:
Thanks!

kerrermanisNL:
Hi, when can we expect this in an official release?

AngellusMortis:
Bump. Are there any plans to release a new version yet? It appears the v1 API is no longer working.

kerrermanisNL:
@angellusmortis you can find a work around here: https://github.com/python-social-auth/social-core/issues/325

PR #325, opened 2019-03-01
Fix linked no firstname lastname key error

kerrermanisNL:
In our production environment we encounter users without a firstname/lastname set. Don't know how this is
possible but it would be nicer if this were handled more gracefully. So instead of breaking just retrieve
an empty string in both cases.

Any feedback would be greatly appreciated, including if this is a good idea or not ;)

### DIFF
--- a/social_core/backends/linkedin.py
+++ b/social_core/backends/linkedin.py
@@ -17,8 +17,8 @@ class BaseLinkedinAuth(object):
     def get_user_details(self, response):
         """Return user details from Linkedin account"""
         fullname, first_name, last_name = self.get_user_names(
-            first_name=response['firstName'],
-            last_name=response['lastName']
+            first_name=response.get('firstName', ''),
+            last_name=response.get('lastName', '')
         )
         email = response.get('emailAddress', '')
         return {'username': first_name + last_name,


kerrermanisNL:
Hi there, could someone have a look at this?

omab:
@kerrermanisNL, can you verify if this is still valid with the new API v2 change merged from #322 ?

kerrermanisNL:
Hi Oman, I'll probably have time to look it the coming week or after that. I'll let you know, sorry for the late response!

kerrermanisNL:
@omab currently it's kind of hard to test this since, AFAICT, there's no official release which contains this fix yet. I'm a little averse to deploying a specific commit or master branch on production servers. Is there any information for when a new release will be deployed?

tuky:
i am currently testing the latest `master` with our test app. i will report back (first try ran into this `KeyError`, too and now I am trying to configure the app as now proposed in https://github.com/python-social-auth/social-docs/pull/55/files)

tuky:
ok, so we now always run into this bug, it seems. our settings are

```python
SOCIAL_AUTH_LINKEDIN_OAUTH2_SCOPE = ['r_liteprofile', 'r_emailaddress']
SOCIAL_AUTH_LINKEDIN_OAUTH2_FIELD_SELECTORS = ['emailAddress']

SOCIAL_AUTH_LINKEDIN_OAUTH2_EXTRA_DATA = [
    ('id', 'id'),
    ('firstName', 'first_name'),
    ('lastName', 'last_name'),
    ('emailAddress', 'email_address'),
]
```

debugging, we can see, that `get_user_details` is being called with

```python
{
'access_token': [Filtered], 
'emailAddress': [Filtered], 
'expires_in': 5183999
}
```
for the `response` variable.

tuky:
@omab I somehow suspect `user_details_url`:

```python
        fields_selectors = list(set(['id', 'firstName', 'lastName'] +
                                    self.setting('FIELD_SELECTORS', [])))
        # user sort to ease the tests URL mocking
        fields_selectors.sort()
```

it includes `emailAddress` from above configuration, but that should actually not be valid anymore for the `/me` end point in `v2`.

tuky:
ok, that is not it. removing `SOCIAL_AUTH_LINKEDIN_OAUTH2_FIELD_SELECTORS` completely makes the `response` variable

```python
{
'access_token': [Filtered], 
'expires_in': 5183999
}
```

maybe it has to do with old v1 access tokens being used now, when requesting v2.

tuky:
You can ignore what I wrote till now, because

```
social-auth-app-django==3.1.0
-e git+https://github.com/python-social-auth/social-core@d3d9c81bdc8b873ff5610e853dc7e9c8f1133e35#egg=social-core-master
```
still also installs `python-social-auth==3.1.0` because it's a dependency of `social-auth-app-django` and then the stable package overrides the editable dev package.

tuky:
ok, i can confirm, that flipping the order in the requirements makes everything work in `v2`.

kerrermanisNL:
We're now also running the latest master commit in production and can also confirm this solves the issue. Logging in with LinkedIn is possible again.

awachat:
v3.2.0 is released to PyPi.

PR #330, opened 2019-03-07
Replace deprecated Google+ API usage in GoogleOpenIdConnect

balsdorf:
GoogleOpenIdConnect was still using the deprecated Google+ API. This PR updates the URL to be the currently supported version and adds a changelog entry.


omab:
Thanks!

Issue #332, opened 2019-03-08
Add backend - Okta


selten:
Add Okta as a backend. Okta is an identity provider. https://www.okta.com/

selten:
See pull request #333

PR #333, opened 2019-03-08
Add Okta backend - Issue #332

selten:
Adds a backend for Okta.


social_core/backends/okta.py
```diff
@@ -0,0 +1,117 @@
...
+                k = key
+                break
+            except ExpiredSignatureError:
+                k = key
+                break
+            except JWTClaimsError as error:
+                raise AuthTokenError(self, str(error))
+            except JWTError as e:
+                if k is None and client_id == 'a-key':
+                    k = self.get_jwks_keys()[0]
```
omab: Can you get rid off this code which is only for the tests to run?

RyanSiu1995:
Any update for this pull request?

selten:
The normal validation of the token provided by Okta seems to fail.
The validation must be fixed and stable (it isn't in the current state). After that, the pull request is ready.

I'll try to have another stab at it soon to get the validation working.

selten:
Turned out that the validation had an issue with Okta recommending to use `/oauth2/default/` as base path. Using `/oauth2/` as base path actually makes validation work all along and make the tests pass with no issue.

selten:
@omab I have resolved that issue, would it be possible to review it again?

social_core/backends/okta.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+import sys
+if sys.version_info[0] < 3:
+    import urlparse as urlparse
+else:
+    import urllib.parse as urlparse
+
+import requests
+from six.moves.urllib.parse import urljoin
+from jose import jwk, jwt, jws
```
sdementen: I do not see the jose imports being use in the code below. Are they necessary ?

social_core/backends/okta.py
```diff
@@ -0,0 +1,79 @@
...
+Okta OAuth2 and OpenIdConnect:
+    https://python-social-auth.readthedocs.io/en/latest/backends/okta.html
+"""
+import sys
+if sys.version_info[0] < 3:
+    import urlparse as urlparse
+else:
+    import urllib.parse as urlparse
+
+import requests
```
sdementen: requests is not use in this module, is the import necessary?

social_core/backends/okta.py
```diff
@@ -0,0 +1,79 @@
...
+from six.moves.urllib.parse import urljoin
+from jose import jwk, jwt, jws
+from jose.jwt import JWTError, JWTClaimsError, ExpiredSignatureError
+from jose.exceptions import JWSSignatureError
+
+from ..utils import append_slash
+from .oauth import BaseOAuth2
+from .open_id_connect import OpenIdConnectAuth
```
sdementen: importing the OpenIdConnectAuth may fail if the extra option "openidconnect" hasn't been installed (https://python-social-auth.readthedocs.io/en/latest/installing.html#using-the-extras-options).
Could the import be in a try:except: block so that if it fails, the class `OktaOpenIdConnect` is not defined ?

selten: I implemented the OpenIdConnect into its own file, like its done with the Google OpenIdConnect.

social_core/backends/okta.py
selten: I've reorganised the imports to be more efficient.

social_core/backends/okta.py
selten: I've reorganised the imports to be more efficient.

social_core/backends/okta.py
```diff
@@ -0,0 +1,59 @@
...
+        return self.get_json(self._url('/.well-known/openid-configuration?client_id='+self.setting('KEY')))
+
+
+class OktaOAuth2(OktaMixin, BaseOAuth2):
+    """Okta OAuth authentication backend"""
+    name = 'okta-oauth2'
+    REDIRECT_STATE = False
+    ACCESS_TOKEN_METHOD = 'POST'
+    SCOPE_SEPARATOR = ' '
+
```
sdementen: I also had to add the 
`    ID_KEY = "preferred_username"`
to get the UID (via the get_user_id() method) properly set in django social.

sdementen: not user if generic or just for my use case...
and maybe openidconnect need the same ==> move it to the OktaMixin

selten: I just checked this using the flask example from: https://github.com/python-social-auth/social-examples. I don't get this same issue.

selten: > not user if generic or just for my use case...
> and maybe openidconnect need the same ==> move it to the OktaMixin

In any case this would be fine under OktaOAuth2 as the OpenIdConnect class inherits from the OAuth2 class.

selten: I've implemented it anyway.

sdementen: tx! 
to understand the thing, what do you get as get_user_id() when not having ID_KEY defined?

selten: I didn't use this personally. I tested it on the example Flask application.

selten:
@omab could you please review this again?

Tiagojdferreira:
@sdementen @omab @selten what is the status on this PR? Soon I will have to integrate Okta support in my app and I wanted to know if I should copy this logic on my end or if I can wait for it to be merged here

selten:
@Tiagojdferreira it's functional and ready to be merged from my perspective. It's up to @omab to review it and merge it.

social_core/backends/okta.py
```diff
@@ -0,0 +1,60 @@
+"""
+Okta OAuth2 and OpenIdConnect:
+    https://python-social-auth.readthedocs.io/en/latest/backends/okta.html
```
Tiagojdferreira: Didn't you forget to add this page?

selten: No. There's a pull request to the documentation open: https://github.com/python-social-auth/social-docs/pull/56.

rohit04apr:
Hi, I am waiting for this functionality to use in my app authentication. When i can expect this to be live?

Tiagojdferreira:
Thansk Omab! 🎉

Issue #279, opened 2018-09-19
Upgrade facebook from 2.9 to 3.2


dmpe:
Hi, 
facebook has upgraded their login api to 3.1 which has brough quite some changes. 

The impacted backend:
https://github.com/python-social-auth/social-core/blob/master/social_core/backends/facebook.py

The website with changelog:
https://developers.facebook.com/docs/facebook-login/changelog

Major change is: **New scopes and `Additional Review Required for Certain Permissions`**
- [x] by default & without any app review, one can only access 3 fields: `name`, `email`, `profile_picture`

dmpe:
Just to make it clear: It seems to work with the current version.

tisdall:
In 3.0 it says they stop using `user_id`.  Doesn't that break things?

tisdall:
I just tested it myself with 3.1 and it seems to be working correctly.  The `user_id` that's provided is already the "third party id" and not the Facebook one.  It no longer says `user_id` in the list of granted scopes, but I don't really understand how that affects things.

tisdall:
Sorry... I should say I changed the API version setting in developers.facebook.com , but didn't change the `API_VERSION` in the code.  The code specifically calls the 2.9 endpoint, but the setting may upgrade it any way.

dmpe:
Hi, 
That was quick :) , I was just preparing to post this as well. 
I have also changed the API version to 3.1 and that was basically it. 
....
as I understand from this [FB page](https://developers.facebook.com/docs/facebook-login/permissions/?locale=en_US#reference-default), `user_id` was replaced with `default` to encompass those former fields.
Being a novice with Django and social login in general, for my app [b40re.tk](b40re.tk) which I host on Heroku and where FB [login is currently being in the "implementation"](https://github.com/dmpe/django-wohn/tree/master/userMng), what I get from extra data

```
[...]

PR #337, opened 2019-03-22
Update Facebook to Graph API 3.2

dmpe:
Shall close https://github.com/python-social-auth/social-core/issues/279


dmpe:
@ibolit  @tisdall @omab Hi, please could you test this new facebook update. 
It update Graph api to 3.2 and also adds profile picture in the return. 

```
sudo pip3 install git+https://github.com/dmpe/social-core.git@bug/facebook
```

Thank you

ibolit:
Installing the patch to test it

social_core/backends/facebook.py
```diff
@@ -59,6 +58,7 @@ def get_user_details(self, response):
         )
         return {'username': response.get('username', response.get('name')),
                 'email': response.get('email', ''),
+                'profile_picture': response.get('profile_picture', ''),
```
tisdall: I don't think this should be added.  `get_user_details` is just to translate each authentication services response into a unified format.  The closest I could get to docs, though, was this:  https://github.com/omab/python-social-auth/issues/7#issuecomment-18087704

social_core/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '3.1.0'
+__version__ = '3.1.1'
```
tisdall: Not a big deal, but you should leave library versioning to the package maintainer.  They may want to include something that's a breaking change and then no 3.1.1 will ever be released, or they may release a 3.1.1 before ever merging your PR.

dmpe:
@tisdall thanks for the feedback!

tisdall:
So fast!  :)

I looked through the API docs and I don't see anything mentioned about OAuth in the changelogs so hopefully there's nothing else to change.

ibolit:
We have tested it, everything seems to work just fine.

ibolit:
I hate to be a nuisance, but can anyone merge this request?

dmpe:
cc @omab

tisdall:
To be clear, the current v2.9 API will stop working July 18, 2019 according to https://developers.facebook.com/docs/graph-api/changelog/

omab:
Thanks!

tisdall:
@ibolit @dmpe this is now released

Issue #338, opened 2019-03-25
Add backend - ID4me


andreeadima:
Add ID4me as a backend. ID4me is a new standard based on Open ID Connect.

Backend developed at [Cloudfest Hackathon 2019](https://www.cloudfest.com/hackathon)

PR #339, opened 2019-03-25
#338 Add ID4me backend

andreeadima:


social_core/backends/id4me.py
```diff
@@ -0,0 +1,270 @@
...
+            if header['kid'] == key['kid']:
+                if 'alg' not in key:
+                    key['alg'] = 'RS256' if key['kty'] == 'RSA' else 'ES256'
+                return key
+
+    def auth_complete(self, *args, **kwargs):
+        self.validate_state()
+        identity = self.strategy.session_get(self.name + '_identity')
+        openid_configuration = self.get_identity_record(identity)
+        if 'clp' not in openid_configuration:
```
pawel-kow: RP shall never need to read out 'clp', as the user-info delegation to Agent shall rely on distributed claims

social_core/backends/id4me.py
```diff
@@ -0,0 +1,270 @@
...
+        except ExpiredSignatureError:
+            raise AuthTokenError(self, 'Signature has expired')
+        except JWTClaimsError as error:
+            raise AuthTokenError(self, str(error))
+        except JWTError:
+            raise
+
+    @handle_http_errors
+    def user_data(self, access_token, *args, **kwargs):
+        user_token = requests.get(self.userinfo_url(), headers={
```
pawel-kow: This one requires rewrite to utilize distributed claims.
Essential is that initial access token can be only considered valid for authorities' userinfo, but the one to be used with Agent can be different and will be provided only by distributed claims object.
See also: https://gitlab.com/ID4me/general/issues/25

social_core/backends/id4me.py
```diff
@@ -0,0 +1,270 @@
...
+        super(ID4meBackend, self).__init__(*args, **kwargs)
+
+    def get_identity_record(self, identity):
+        try:
+            response = dns.resolver.query('_openid.' + identity, 'TXT', lifetime=5).response.answer
+        except NXDOMAIN or Timeout:
+            raise AuthUnreachableProvider(self)
+        if not response:
+            raise AuthUnreachableProvider(self)
+        records = response[0]
```
pawel-kow: worth iterating all returned records

social_core/backends/id4me.py
```diff
@@ -0,0 +1,270 @@
...
+        if not records:
+            raise AuthUnreachableProvider(self)
+        record = records[-1].strings[0].decode()
+        return {item.split("=")[0]: item.split("=")[1] for item in record.split(";")}
```
pawel-kow: there shall be a check that record has v=OID1 label set

social_core/backends/id4me.py
```diff
@@ -0,0 +1,270 @@
...
+        except IndexError:
+            return None
+
+    @handle_http_errors
+    def get_key_and_secret(self):
+        iau = self.strategy.session_get(self.name + '_authority')
+        association = self.get_association(iau)
+        if not association:
+            issuer_configuration = self.oidc_config_authority()
+            response = requests.post(issuer_configuration['registration_endpoint'], json={
```
pawel-kow: there shall be some configuration added for policy/logo/tos URLs

social_core/backends/id4me.py
```diff
@@ -0,0 +1,270 @@
...
+        try:
+            return self.strategy.storage.association.get(server_url=issuer)[0]
+        except IndexError:
+            return None
+
+    @handle_http_errors
+    def get_key_and_secret(self):
+        iau = self.strategy.session_get(self.name + '_authority')
+        association = self.get_association(iau)
+        if not association:
```
pawel-kow: there shall be a check, that registration did not expire (client_secret_expires_at)

social_core/backends/id4me.py
```diff
@@ -0,0 +1,270 @@
...
+    EXTRA_DATA = ['sub', 'iss', 'clp']
+    JWT_DECODE_OPTIONS = dict(verify_at_hash=False)
+
+    def __init__(self, *args, **kwargs):
+        super(ID4meBackend, self).__init__(*args, **kwargs)
+
+    def get_identity_record(self, identity):
+        try:
+            response = dns.resolver.query('_openid.' + identity, 'TXT', lifetime=5).response.answer
+        except NXDOMAIN or Timeout:
```
pawel-kow: There shall be a lookup on parent domain in case of failure.
See https://gitlab.com/ID4me/documentation/blob/master/id4ME Technical Specification.adoc 2.1.2
> In case _openid subdomain cannot be resolved, the record for the parent domain SHALL be resolved, by removing the leftmost label from the original ID4me Identifier.

social_core/backends/id4me.py
```diff
@@ -0,0 +1,270 @@
...
+    def get_jwks_keys(self):
+        keys = self.get_remote_jwks_keys()
+        return keys
+
+    def find_valid_key(self, id_token):
+        for key in self.get_jwks_keys():
+            header = jwt.get_unverified_header(id_token)
+            if header['kid'] == key['kid']:
+                if 'alg' not in key:
+                    key['alg'] = 'RS256' if key['kty'] == 'RSA' else 'ES256'
```
pawel-kow: typically there shall be a correlation between `id_token_signed_response_alg` set in the client registration and the signature key. If omitted they default to 'RSA256', so usage of 'ES256' is not correct in this case.

social_core/backends/id4me.py
```diff
@@ -0,0 +1,270 @@
...
+                return key
+
+    def find_agent_valid_key(self, id_token):
+        for key in self.get_agent_keys():
+            header = jwt.get_unverified_header(id_token)
+            if header['kid'] == key['kid']:
+                if 'alg' not in key:
+                    key['alg'] = 'RS256' if key['kty'] == 'RSA' else 'ES256'
```
pawel-kow: typically there shall be a correlation between `userinfo_signed_response_alg` set in the client registration and the signature key. If omitted they default to 'RSA256', so usage of 'ES256' is not correct in this case.

social_core/backends/id4me.py
```diff
@@ -0,0 +1,270 @@
...
+        return self.get_key_and_secret()
+
+    def validate_claims(self, id_token):
+        utc_timestamp = timegm(datetime.datetime.utcnow().utctimetuple())
+
+        if 'nbf' in id_token and utc_timestamp < id_token['nbf']:
+            raise AuthTokenError(self, 'Incorrect id_token: nbf')
+
+        # Verify the token was issued in the last 10 minutes
+        iat_leeway = self.setting('ID_TOKEN_MAX_AGE', self.ID_TOKEN_MAX_AGE)
```
pawel-kow: Why this arbitrary check for 10 minutes? There is `exp` claim which shall be respected.

PR #340, opened 2019-03-27
bugfix: github backend need `SEND_USER_AGENT`

laixintao:
From the [document](https://developer.github.com/v3/#user-agent-required):

> All API requests MUST include a valid User-Agent header. Requests with no User-Agent header will be rejected. We request that you use your GitHub username, or the name of your application, for the User-Agent header value. This allows us to contact you if there are problems.

I got 403 forbidden when login with github.

> Your access to this site has been restricted.

image

It turns out that if add User-Agent then everything is fine.


omab:
Thanks

PR #346, opened 2019-04-01
Use hmac.compare_digest for value comparison

codingjoe:
`hmac.compare_digest` is available in both Python 2/3
and provides time constant comparison of ascii and binary
values. It is safe for cryptographic use to prevent timing
attacks.


codingjoe:
Sorry for the commit mass. I did this change in the browser. Let me know if you want me to squash it first. Besides that, this should be read to go.

omab:
Thanks!

Issue #263, opened 2018-07-20
user_details pipeline fails to update changed user data


LiamK:
The user_details pipeline does not update the local user when the user details from the OAuth provider have been changed.  In the following code from user_details() in user.py, line 98 is testing for a truthy value for current_value, but it shouldn't since that means that changed details will not be updated:
```
        # Check https://github.com/omab/python-social-auth/issues/671
        current_value = getattr(user, name, None)
        if current_value or current_value == value:
            continue
```
To test this I changed my name on the OAuth provider, logged out of the application, and logged in again.  The name field had not been updated.  I tracked it down to the code above.

I substituted:
```
        try:
            current_value = getattr(user, name)
            if current_value == value:
                continue
        except AttributeError:
            continue
```
And it works correctly for my purposes.  Note that this can possibly update a field value to None or empty string (whatever is in the details).  I think that is the correct behavior.

This passes all the tests, but there isn't a test that demonstrates the incorrect behavior, and I'm not quite sure how to write one.  It should be quite simple, if someone wants to give me some tips.

BTW, email is hard-coded as a protected field.  I'm wondering if that is correct, because I can envision the user changing their email at some point, and wanting that field to be updated.

LiamK:
Actually, it is even simpler, since the check for missing attribute is already being done above.

```
        current_value = getattr(user, name)
        if current_value == value:
            continue
```

gnalbandian:
I have confirmed this issue, but haven't got fixed with your suggested solution. Will test better tomorrow and come back. I did the tests with github.

sposs:
I believe this requires a pull request...

sposs:
I have create my own copy, where I implemented your check. It now doing the 'right' thing. I'm actually curious as to what drove the current implementation...

dadon-david:
Hi Everyone,

I created a PR: https://github.com/python-social-auth/social-core/pull/348

Please upvote 😃

dadon-david:
My PR seems to failed on one test.
I will check that this weekend,

Meanwhile Does anybody have an idea on why it is failing ?

Thank you 

cc @omab :)

nicholasamorim:
Pretty obvious case for backends such as SAML, where email can be transient. Can we have some review on the PR #348? - cc @omab

nicholasamorim:
An alternative way to forking this repo is to replace the function in the pipeline:

Put this in your settings:

```python
SOCIAL_AUTH_PIPELINE = (
    # All default except the last
[...]

PR #348, opened 2019-04-04
user_details bugfix - Updating user model attributes

dadon-david:
user_details currently does not update the user model attributes because of a bug in the condition that check if an update is necessary (line 98)

Indeed, if any value was previously set, the user would not be updated.

This pull request fix that and now the user model attributes correctly get updated.

Thank you very much

### DIFF
--- a/social_core/pipeline/user.py
+++ b/social_core/pipeline/user.py
@@ -95,7 +95,7 @@ def user_details(strategy, details, user=None, *args, **kwargs):
 
         # Check https://github.com/omab/python-social-auth/issues/671
         current_value = getattr(user, name, None)
-        if current_value or current_value == value:
+        if current_value == value:
             continue
 
         changed = True


dadon-david:
Related to: https://github.com/python-social-auth/social-core/issues/263

LiamK:
Thanks!

LiamK:
Is this going to be merged?

dadon-david:
Only one test failed and I don't understand why.
@omab can you help us on this case, ? Thank you very much

SpoonMeiser:
You should also remove the comment that's no longer helpful in understanding the code.

Binnette:
Hi @dadon-david, can I suggest you to close your PR?
There is an "improved" version of your PR here: #432 
Have a nice day.

dadon-david:
Closing this PR. There is a better version here: https://github.com/python-social-auth/social-core/pull/432

PR #350, opened 2019-04-10
Add email_verified to user details

james-emerton:
It's useful to know if the user's email address has been verified. In my case, I'm using this field in a pipeline to require a verified email before merging to an existing user.


omab:
Merged, thanks!

PR #353, opened 2019-04-15
Prepare for Shopify SDK v5

Maks3w:
Shopify v5 introduces the mandatory requirement to provide a version when setup.

https://github.com/Shopify/shopify_python_api#-breaking-change-notice-for-version-500-

### DIFF
--- a/social_core/backends/shopify.py
+++ b/social_core/backends/shopify.py
@@ -42,7 +42,7 @@ class ShopifyOAuth2(BaseOAuth2):
         extra_data field"""
         data = super(ShopifyOAuth2, self).extra_data(user, uid, response,
                                                      details, *args, **kwargs)
-        session = self.shopifyAPI.Session(self.data.get('shop').strip())
+        session = self.shopifyAPI.Session(self.data.get('shop').strip(), version='2019-04')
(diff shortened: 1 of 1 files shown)


omab:
@Maks3w code looks good to me, but what do you think about letting the version be defined on a setting (`SHOPIFY_API_VERSION`?) with the `2019-04` value as a default?

Maks3w:
I think that value is shouldn't be customizable. PSA calls Shopify as it expected to be and forward compatibility is unknown.

omab:
@Maks3w looks like they support also an `unstable` as value for `version`, this might come handy for testing future versions.

Maks3w:
I suppose Shopify don't get mad and remove `2019-04` without previous notice.

`unstable` make sense for someone who wants to use the SDK in a low level mode.

IMO this is not the case. It's better to lock to a known version. If someone wants in the future to add new code in PSA project feel free to change it, also still be possible for the developer to use different API versions even if PSA is locked to a particular one.

omab:
The problem with a lock down version is that it will demand a whole new PyPi release just to bump this up, while a setting will let users use the new version instead of having they wait.

PR merged, thanks!

PR #359, opened 2019-04-27
Add OpenStack Backends

eumel8:
OpenStack Community is used to OpenID services:
http://openstackid-dev.openstack.org/
http://openstackid.org/
Both are included.


omab:
Thanks!

Issue #335, opened 2019-03-20
GoogleOAuth2 access token data fetch URL


sposs:
In the Google Sign-in doc for iOS: https://developers.google.com/identity/sign-in/ios/backend-auth, they mention that to get the user data, one should use https://oauth2.googleapis.com/tokeninfo. In the current social-core code, https://github.com/python-social-auth/social-core/blob/master/social_core/backends/google.py#L49, www.googleapis.com (and auth headers) is used. This latter one is restricted in number of calls (if not using the API + services in google), while the one in the iOS doc does not seem to be restricted.

PR #363, opened 2019-05-13
Use unrestricted API for user info

sposs:
Given #335, I propose to change the URL. The default one is restricted in the number of queries per day.


sposs:
The tests are failing, but I can't understand why...

sposs:
This pull request solves one issue (connection using iOS SDK with returns id_token, not access_token), but breaks the login via the /complete/google-oauth endpoint, because this uses the access_token, not the same as id_token. 
Not sure how to address this in the end... Any idea?

sposs:
Solution: try both...

sposs:
hello,
I've merged with Master, in the hope to have this merged eventually. Can the checks be enabled?

PR #367, opened 2019-05-29
switch to pytest

davidszotten:
Initial pr that just switches the test runner to pytest, leaving all
tests. All tests are found and pass. We can then gradually start using
pytest features.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,10 +3,5 @@ max-line-length = 119
 # Ignore some well known paths
 exclude = .venv,.tox,dist,doc,build,*.egg,db/env.py,db/versions/*.py,site
 
-[nosetests]
-verbosity=2
-with-coverage=1
-cover-erase=1
-cover-package=social_core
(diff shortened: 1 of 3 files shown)


omab:
@davidszotten, I don't mind updating the test suite, but what's the reasoning for the change? Could you enlighten me, please?

davidszotten:
from https://nose.readthedocs.io/en/latest/
>Note to Users
Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2.

pytest is an increasingly popular, and very nice testing framework, requiring much less boilerplate than unittest. see https://docs.pytest.org/en/latest/ for a brief overview

as i mentioned, pytest happily collects and runs unittest tests, which enables an easy and gradual transition

omab:
@davidszotten, thanks for the clarification, the change is very welcome. Thanks!

Issue #365, opened 2019-05-28
Partial Prepare function under partial pipelines fails for model with custom primary key field


kimbugp:
### Description
When using partial pipeline such as `social_core.pipeline.mail.mail_validation` with a user model having the primary key other than `id`, the function `partial_prepare` https://github.com/python-social-auth/social-core/blob/master/social_core/pipeline/utils.py#L17 raises an `attribute error`

### Steps to Reproduce
- Create a model with a primary key field name other than `id`
-  Add `social_core.pipeline.mail.mail_validation` to the `SOCIAL_AUTH_PIPELINE` 
- Perform a request to add a new user.
 
### Earliest Version Affected
- social-auth-core v3.1.0

### Error
```
/social_core/backends/base.py", line 80, in authenticate
return self.pipeline(pipeline, *args, **kwargs)

/social_core/backends/base.py", line 83, in pipeline
    out = self.run_pipeline(pipeline, pipeline_index, *args, **kwargs)

/social_core/backends/base.py", line 113, in run_pipeline
    result = func(*args, **out) or {}

/social_core/pipeline/partial.py", line 28, in wrapper
    *args, **kwargs)
 /social_core/pipeline/utils.py", line 26, in partial_prepare
    'user': user and user.id or None,
AttributeError: 'User' object has no attribute 'id'
```

### Possible Implementation

- Use the attribute `pk` instead of `id` since it's independent of the actually primary key

PR #368, opened 2019-05-29
Check for id attribute in user object 

kimbugp:
### What does this PR do
- check if id attribute exists the user object  before assigning it to `kwargs`
- fixes Issue #365 
### Related issues 
- #365


omab:
Thanks!

PR #371, opened 2019-06-10
Ready to prevent API deprecation for backends.naver

jeyraof:
for backends.naver


jeyraof:
I merge commits by fixing tests. And reopen it

PR #388, opened 2019-08-18
Fix auto remove unicode characters in username

tinhien11:
Unicode in username should be covert to ASCII is better than remove

### DIFF
--- a/requirements-python2.txt
+++ b/requirements-python2.txt
@@ -1,2 +1,3 @@
 python-openid>=2.2.5
+unidecode>=1.1.1
 -r requirements-base.txt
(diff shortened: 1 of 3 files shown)


omab:
Love it, thanks!

ryanpetrello:
Are you all aware that Unidecode is GPLv2?

omab:
@tinhien11, this change had to be reverted due to license incompatibilities between `python-social-auth` (BSD) and `unidecode` (GLPv2+). I've added an entry in the `Use Cases` section of the docs on how to accomplish something similar just with configuration: https://python-social-auth.readthedocs.io/en/latest/use_cases.html#improve-unicode-cleanup-from-usernames

PR #394, opened 2019-09-23
Twitch API v5 compatibility.

onursa:


batuzunoglu:
teşekkürler twitch login'de hata alıyordum sorunum düzeldi

batuzunoglu:
Thank You! I was having trouble login to twitch. But the problem is solved.

DorkmasterFlek:
@omab Will this be released soon? The old Twitch API has been shut down for several months and Twitch authentication does not work at all with the current version of this library released back in May. I've been installing it from the GitHub repo directly as a workaround, but this is not proper for production releases.

PR #399, opened 2019-10-01
Add support for SurveyMonkey and HubSpot backends

svvitale:


svvitale:
I just pushed one additional commit to also include the HubSpot backend.  I'm happy to do this as two separate PRs if you'd like.

PR #404, opened 2019-10-22
Mailru backend issue 403 - close403

pulysak:
There is already backend for mailru:
https://python-social-auth.readthedocs.io/en/latest/backends/mailru.html
https://github.com/python-social-auth/social-core/blob/master/social_core/backends/mailru.py

which basically use https://my.mail.ru/ - old social network for authentication.
I think it is a good idea to do a new backend for the new mail.ru Oauth - https://oauth.mail.ru/docs


firewut:
@omab, what do you think about it ?:)

firewut:
@omab, thank you very much! :)

PR #409, opened 2019-11-17
Implement Sign In With Apple backend

### LINKED ISSUE #370: Add backend - Apple
Add support for Sign in with Apple:

https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple

grazor:
Hi!

I've implemented Sign In With Apple backend and tested it with both web and mobile applications. Hope this might be useful.

Resolves #370.

### DIFF
--- a/requirements-base.txt
+++ b/requirements-base.txt
@@ -3,3 +3,4 @@ oauthlib>=1.0.3
 requests-oauthlib>=0.6.1
 six>=1.10.0
 PyJWT>=1.4.0
+cryptography>=1.4
--- a/social_core/backends/apple.py
+++ b/social_core/backends/apple.py
@@ -1,0 +1,131 @@
+"""
+Sign In With Apple authentication backend.
+
+Docs:
+    * https://developer.apple.com/documentation/signinwithapplerestapi
+    * https://developer.apple.com/documentation/signinwithapplerestapi/tokenresponse
+
+Settings:
+    * `TEAM` - your team id;
+    * `KEY` - your key id;
+    * `CLIENT` - your client id;
+    * `SECRET` - your secret key;
+    * `SCOPE` (optional) - e.g. `['name', 'email']`;
+    * `EMAIL_AS_USERNAME` - use apple email is username if set, use apple id otherwise.
+    * `AppleIdAuth.TOKEN_TTL_SEC` - time before JWT token expiration, seconds.
+"""
+
+import json
+import time
+
+import jwt
+from jwt.algorithms import RSAAlgorithm
+from jwt.exceptions import PyJWTError
+
+from social_core.backends.oauth import BaseOAuth2
+from social_core.exceptions import AuthCanceled
+
+
+class AppleIdAuth(BaseOAuth2):
+    name = "apple-id"
+
+    JWK_URL = "https://appleid.apple.com/auth/keys"
+    AUTHORIZATION_URL = "https://appleid.apple.com/auth/authorize"
+    ACCESS_TOKEN_URL = "https://appleid.apple.com/auth/token"
+    ACCESS_TOKEN_METHOD = "POST"
+    RESPONSE_MODE = None
+
+    ID_KEY = "sub"
+    TOKEN_KEY = "id_token"
+    STATE_PARAMETER = True
+    REDIRECT_STATE = False
+
+    TOKEN_AUDIENCE = "https://appleid.apple.com"
+    TOKEN_TTL_SEC = 6 * 30 * 24 * 60 * 60
+
+    def auth_params(self, *args, **kwargs):
+        """Apple requires to set `response_mode` to `form_post` if `scope` parameter is passed."""
+        params = super(AppleIdAuth, self).auth_params(*args, **kwargs)
+        if self.RESPONSE_MODE:
+            params["response_mode"] = self.RESPONSE_MODE
+        elif self.get_scope():
+            params["response_mode"] = "form_post"
+        return params
+
+    def get_private_key(self):
+        """
+        Return contents of the private key file.
+        Override this method to provide secret key from another source if needed.
+        """
+        return self.setting("SECRET")
+
+    def generate_client_secret(self):
+        now = int(time.time())
+        client_id = self.setting("CLIENT")
+        team_id = self.setting("TEAM")
+        key_id = self.setting("KEY")
+        private_key = self.get_private_key()
+
+        headers = {"kid": key_id}
+        payload = {
+            "iss": team_id,
+            "iat": now,
+            "exp": now + self.TOKEN_TTL_SEC,
+            "aud": self.TOKEN_AUDIENCE,
+            "sub": client_id,
+        }
+
+        return jwt.encode(payload, key=private_key, algorithm="ES256", headers=headers)
+
+    def get_key_and_secret(self):
+        client_id = self.setting("CLIENT")
+        client_secret = self.generate_client_secret()
+        return client_id, client_secret
+
+    def get_apple_jwk(self):
+        keys = self.get_json(url=self.JWK_URL).get("keys")
+        if not isinstance(keys, list) or not keys:
+            raise AuthCanceled("Invalid jwk response")
+
+        return json.dumps(keys.pop())
+
+    def decode_id_token(self, id_token):
+        """Decode and validate JWT token from apple and return payload including user data."""
+        if not id_token:
+            raise AuthCanceled("Missing id_token parameter")
+
+        public_key = RSAAlgorithm.from_jwk(self.get_apple_jwk())
+        try:
+            decoded = jwt.decode(id_token, key=public_key, audience=self.setting("CLIENT"), algorithm="RS256",)
+        except PyJWTError:
+            raise AuthCanceled("Token validation failed")
+
+        return decoded
+
+    def get_user_details(self, response):
+        name = response.get("name") or {}
+        fullname, first_name, last_name = self.get_user_names(
+            fullname="", first_name=name.get("firstName", ""), last_name=name.get("lastName", ""),
+        )
+
+        email = response.get("email", "")
+        apple_id = response.get(self.ID_KEY, "")
+        user_details = {
+            "first_name": first_name,
+            "last_name": last_name,
+            "email": email,
+        }
+        if email and self.setting("EMAIL_AS_USERNAME"):
+            user_details["username"] = email
+        if apple_id and not self.setting("EMAIL_AS_USERNAME"):
+            user_details["username"] = apple_id
+
+        return user_details
+
+    def do_auth(self, access_token, response, *args, **kwargs):
+        jwt_string = kwargs.pop("jwt_string", None) or response.get(self.TOKEN_KEY)
+        if not jwt_string:
+            raise AuthCanceled("Missing id_token parameter")
+
+        decoded_data = self.decode_id_token(jwt_string)
+        return super(AppleIdAuth, self).do_auth(access_token, response=decoded_data, *args, **kwargs)
--- a/social_core/backends/apple.py
+++ b/social_core/backends/apple.py
@@ -21,7 +21,6 @@ import time
 import jwt
 from jwt.algorithms import RSAAlgorithm
 from jwt.exceptions import PyJWTError
-
 from social_core.backends.oauth import BaseOAuth2
 from social_core.exceptions import AuthCanceled
 
@@ -122,8 +121,9 @@ class AppleIdAuth(BaseOAuth2):
 
         return user_details
 
-    def do_auth(self, access_token, response, *args, **kwargs):
-        jwt_string = kwargs.pop("jwt_string", None) or response.get(self.TOKEN_KEY)
+    def do_auth(self, access_token, *args, **kwargs):
+        response = kwargs.pop('response', None) or {}
+        jwt_string = response.get(self.TOKEN_KEY) or access_token
         if not jwt_string:
             raise AuthCanceled("Missing id_token parameter")
 
--- a/social_core/backends/apple.py
+++ b/social_core/backends/apple.py
@@ -11,7 +11,7 @@ Settings:
     * `CLIENT` - your client id;
     * `SECRET` - your secret key;
     * `SCOPE` (optional) - e.g. `['name', 'email']`;
-    * `EMAIL_AS_USERNAME` - use apple email is username if set, use apple id otherwise.
+    * `EMAIL_AS_USERNAME` - use apple email is username is set, use apple id otherwise.
     * `AppleIdAuth.TOKEN_TTL_SEC` - time before JWT token expiration, seconds.
 """
 
@@ -21,6 +21,7 @@ import time
 import jwt
 from jwt.algorithms import RSAAlgorithm
 from jwt.exceptions import PyJWTError
+
 from social_core.backends.oauth import BaseOAuth2
 from social_core.exceptions import AuthCanceled
 
@@ -82,21 +83,26 @@ class AppleIdAuth(BaseOAuth2):
         return client_id, client_secret
 
     def get_apple_jwk(self):
+        """Returns an iterable of apple jwk strings."""
         keys = self.get_json(url=self.JWK_URL).get("keys")
         if not isinstance(keys, list) or not keys:
             raise AuthCanceled("Invalid jwk response")
 
-        return json.dumps(keys.pop())
+        return (json.dumps(key) for key in keys)
 
     def decode_id_token(self, id_token):
         """Decode and validate JWT token from apple and return payload including user data."""
         if not id_token:
             raise AuthCanceled("Missing id_token parameter")
 
-        public_key = RSAAlgorithm.from_jwk(self.get_apple_jwk())
-        try:
-            decoded = jwt.decode(id_token, key=public_key, audience=self.setting("CLIENT"), algorithm="RS256",)
-        except PyJWTError:
+        for jwk_string in self.get_apple_jwk():
+            public_key = RSAAlgorithm.from_jwk(jwk_string)
+            try:
+                decoded = jwt.decode(id_token, key=public_key, audience=self.setting("CLIENT"), algorithm="RS256",)
+                break
+            except PyJWTError:
+                pass
+        else:
             raise AuthCanceled("Token validation failed")
 
         return decoded
--- a/social_core/tests/backends/test_apple.py
+++ b/social_core/tests/backends/test_apple.py
@@ -1,0 +1,54 @@
+import json
+
+from .oauth import OAuth2Test
+
+try:
+    from unittest.mock import patch
+except ImportError:
+    from mock import patch
+
+
+TEST_KEY = """
+-----BEGIN EC PRIVATE KEY-----
+MHcCAQEEIKQya8aIoeoOLeThk7Ad/lLyAo2fTp9IuhIpy2CivH/qoAoGCCqGSM49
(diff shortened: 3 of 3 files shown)


abozhilov:
Using this I got: `do_auth() missing 1 required positional argument: 'response'`

grazor:
> Using this I got: `do_auth() missing 1 required positional argument: 'response'`

Could you please provide a stack trace? How do you use this backend? What version of social-auth-core do you use?

abozhilov:
Thanks for your reply. 
social-auth-core==3.2.0

I use it in conjunction with: `django-rest-framework-social-oauth2` 

The traceback is following: 
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.7/dist-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.7/dist-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.7/dist-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/django/utils/decorators.py", line 45, in _wrapper
    return bound_method(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/braces/views/_forms.py", line 24, in dispatch
    return super(CsrfExemptMixin, self).dispatch(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/rest_framework/views.py", line 497, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.7/dist-packages/rest_framework/views.py", line 457, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.7/dist-packages/rest_framework/views.py", line 468, in raise_uncaught_exception
    raise exc
  File "/usr/local/lib/python3.7/dist-packages/rest_framework/views.py", line 494, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/rest_framework_social_oauth2/views.py", line 70, in post
    url, headers, body, status = self.create_token_response(request._request)
  File "/usr/local/lib/python3.7/dist-packages/oauth2_provider/views/mixins.py", line 123, in create_token_response
    return core.create_token_response(request)
  File "/usr/local/lib/python3.7/dist-packages/oauth2_provider/oauth2_backends.py", line 138, in create_token_response
    headers, extra_credentials)
  File "/usr/local/lib/python3.7/dist-packages/oauthlib/oauth2/rfc6749/endpoints/base.py", line 116, in wrapper
    return f(endpoint, uri, *args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/rest_framework_social_oauth2/oauth2_endpoints.py", line 60, in create_token_response
    request, self.default_token_type)
  File "/usr/local/lib/python3.7/dist-packages/oauthlib/oauth2/rfc6749/grant_types/refresh_token.py", line 60, in create_token_response
    self.validate_token_request(request)
  File "/usr/local/lib/python3.7/dist-packages/rest_framework_social_oauth2/oauth2_grants.py", line 94, in validate_token_request
    user = backend.do_auth(access_token=request.token)
TypeError: do_auth() missing 1 required positional argument: 'response'
``` 
This is what I got after calling convert token endpoint with the following properties:
```
grant_type=convert_token&client_id=<client_id>&client_secret=<client_secret>&backend=apple-id&token=<apple_id_token>
```

The configuration of AppleID is following:
```
SOCIAL_AUTH_APPLE_ID_TEAM = 'team_id'
SOCIAL_AUTH_APPLE_ID_KEY = 'key_id'
SOCIAL_AUTH_APPLE_ID_CLIENT = 'service.com.id'
SOCIAL_AUTH_APPLE_ID_SECRET = '''-----BEGIN PRIVATE KEY-----
.....
-----END PRIVATE KEY-----'''
```

The backend is loaded as following:
```
AUTHENTICATION_BACKENDS = (
    # Facebook OAuth2
    'social_core.backends.facebook.FacebookAppOAuth2',
    'social_core.backends.facebook.FacebookOAuth2',
    # Google OAuth2
    'social_core.backends.google.GoogleOAuth2',
    #Twitter OAuth
    'social_core.backends.twitter.TwitterOAuth',
    #AppleID OAuth2
    'users.apple_backend.AppleOAuth2',
    # django-rest-framework-social-oauth2
    'rest_framework_social_oauth2.backends.DjangoOAuth2'
)

```

Any help is appreciated! 
Thank you.

grazor:
Hi @abozhilov !

Thanks for such a detailed information! I've pushed a small fix that should resolve your issue, however I'm not able test it with drf right now. Could you please check if this solution works for you?

bmax329:
do we think this will get merged soon? April is right around the corner, and we'd love to be able to use a built-in solution 😁

social_core/backends/apple.py
```diff
@@ -0,0 +1,131 @@
...
+        public_key = RSAAlgorithm.from_jwk(self.get_apple_jwk())
+        try:
+            decoded = jwt.decode(id_token, key=public_key, audience=self.setting("CLIENT"), algorithm="RS256",)
+        except PyJWTError:
+            raise AuthCanceled("Token validation failed")
+
+        return decoded
+
+    def get_user_details(self, response):
+        name = response.get("name") or {}
```
bmax329: I'm not sure if I'm misusing it, but when I implemented this verbatim in my app, `response` appeared to be the initial oauth response from apple and did not have the user fields.

I ended up changing this so that response was instead defined as

```
response = json.loads(self.data.get('user'))
```

From there, the rest of the name parsing logic matched up.

bmax329: Ah - I had it backwards. `response` is the decoded id_token, which [does not include the name anymore](https://forums.developer.apple.com/thread/124169).

Given email/name all comes back in the initial post (`self.data`), it almost feels like there is little need to do go beyond the `auth_complete` step since the post data with the `code` already includes the user info fields.

XaviTorello:
I was dealing with `invalid grant` responses from the AppleID sign in service for websites:
```
{ 'error': 'invalid_grant' }
```

I proposed a solution in https://github.com/grazor/social-core/pull/1  (https://github.com/XaviTorello/social-core/tree/sign_in_with_apple)

I saw that the final `redirect_uri` param in the payload submitted from [`auth_complete`](https://github.com/python-social-auth/social-core/blob/17d7f66f6f8310dc2ebad86df5f5813721fe55ce/social_core/backends/oauth.py#L385) is `/`.

https://github.com/python-social-auth/social-core/blob/17d7f66f6f8310dc2ebad86df5f5813721fe55ce/social_core/backends/oauth.py#L345-L353

It should be enough for app sign in requests, but for website it should match the original `request_uri` provided in the original request (the one that redirect's to the social-auth handler).

I'm talking about the `auth/authorize` AppleID service, i.e https://appleid.apple.com/auth/authorize?client_id=client.a.is.this&redirect_uri=https://yourserver.com/api/v1/oauth/social/jwt-pair/apple-id/&response_type=code&scope=name+email&response_mode=form_post&state=45

Would be great if anyone can confirm if app sign in requests still works as expected with this change.

sposs:
I can confirm this pull request works like a charm. Can a package manager merge and prepare a release?
 
Thanks for doing the work!

siell23:
+1, works great! 

Only change I locally made was keeping only the first part of the email before the "@" as the username, as otherwise the "@" gets stripped by something later in the pipeline and makes an ugly username.

sposs:
@siell23 this is something you can implement in the pipeline yourself. I don't think you should change the default behavior of this code. For example, in my case, the usernames are not at all related to the emails (especially with apple and the 'Hide your email' feature, where they provide gibberish emails).

siell23:
@sposs Yeah I wanted to at first, we already have some hooks in the pipeline that deal with username-related things, but it looked like the stripping of the `@` happened earlier. I then looked at the existing Google auth code and saw that it keeps only the first bit of the email (https://github.com/python-social-auth/social-core/blob/master/social_core/backends/google.py#L38).

I have to admit though that I didn't try very hard to dig into the code to find out where exactly it happens. At the moment since I've literally just copy-pasted the code from this PR into our codebase, it was easiest to change it there. But once this PR is merged, I can try and find another work around :)

dcmoore-gd:
Does anyone know what the impediment for getting this merged is?  Based on the rest of the repo including the [Contributing Guidelines](https://github.com/python-social-auth/social-core/blob/master/CONTRIBUTING.md) I'm guessing that @omab might be waiting on:

- [ ] Add a meaningful entry to the CHANGELOG.md document.
- [ ] A resolution to @XaviTorello's question above.
- [ ] A corresponding PR in the https://github.com/python-social-auth/social-docs repo to add an appropriate entry to https://github.com/python-social-auth/social-docs/tree/master/docs/backends

I'm happy to help implement or review any of this but don't want to step on @grazor's work!  And again thanks to @grazor for this PR and to omab for python-social-auth!

grazor:
@dcmoore-gd Thanks for detailed explanation of the actions I've missed! I'll resolve it as soon as I can

dcmoore-gd:
@grazor I have no authority or special knowledge just guessing!  And please let everyone here know if you want some help!

XaviTorello:
Related to my question, I found a simple solution.

I proposed a PR to @grazor's to be able to configure if the backend should enforce the requested URL as `redirect_uri` using a new setting `ENFORCE_REDIRECT_URL_WITH_REQUESTED_URI`. See https://github.com/grazor/social-core/pull/1

With this, authorization flows that uses redirects like `auth/authorize` AppleID service (https://appleid.apple.com/auth/authorize?client_id=client.a.is.this&redirect_uri=https://yourserver.com/api/v1/oauth/social/jwt-pair/apple-id/&response_type=code&scope=name+email&response_mode=form_post&state=4214j3ijasd) will also work.

Maybe will be simpler and quicker if this PR is merged, and I'll propose changes directly to the main repo. I've some other improvements and fixes to apply for this backend.

Regarding to the documentation, I submitted https://github.com/python-social-auth/social-docs/pull/68

social_core/backends/apple.py
```diff
@@ -0,0 +1,137 @@
...
+            raise AuthCanceled("Invalid jwk response")
+
+        return (json.dumps(key) for key in keys)
+
+    def decode_id_token(self, id_token):
+        """Decode and validate JWT token from apple and return payload including user data."""
+        if not id_token:
+            raise AuthCanceled("Missing id_token parameter")
+
+        for jwk_string in self.get_apple_jwk():
```
bmax329: So I've been following this PR for a few months and know this iterating retry logic was not here originally. Out of curiosity, what led to this change? Did something change on Apple's side?

XaviTorello: Yesterday Apple started to sign with two new certs. See https://appleid.apple.com/auth/keys

Iterate is not the best approach, but will be great if this PR can be merged, I've two improvements ready to be proposed

sposs:
I've been trying to use this to authenticate from a device: the iOS device obtains a id_token and sends it. Unfortunately, the verification fails always, and I had to put ```verify=False``` in the ```jwt.decode``` step. Any idea why the verification would fail?

grazor:
@sposs Perhaps it happens because of a behavior of `keys` endpoint returning multiple keys @XaviTorello pointed out earlier. Did you try a (dirty?) fix I pushed yesterday?

sposs:
I found out what the problem was: in my tests I was using an id token provided by the iOS app, but I had not checked the expiration date. That thing expires after 10 minutes!
Now, my concern is that the CLIENT that's set in the backend configuration for the web access is different from the one needed to decode the id_token from the app. This is because of course apple does not allow a service ID to be the same as the bundle ID.

sposs:
I did actually implement your fix for the multiple keys. It was needed because the key needed was the second one.

dcmoore-gd:
Oh wow, that's good information @sposs about the certs.  Has anyone found a good resource/wiki/discussion outside of Apple's Dev documents about actually implementing this?  Might be useful to have one specific to the `python-social-auth` use case in any case...

sposs:
Moreover, the app SDK does not follow the same rules as the other OpenID/OAuth systems: they do not allow 'refresh' of the token (as far as we could see): their 'refresh' operation is essentially a new login. So there is no way to easily refresh the token in the app without bugging the user, and therefore (because the token expires after 10 minutes), we cannot use it for a subsequent auth. I had to add a dedicated endpoint that takes as param the apple user ID and looks in the users for the one that's got this social UID and that's it, user is authenticated. It's fairly poor in terms of security, but for our purpose it's ok.If anyone has a better solution, I'm interested!

XaviTorello:
> I found out what the problem was: in my tests I was using an id token provided by the iOS app, but I had not checked the expiration date. That thing expires after 10 minutes!
> Now, my concern is that the CLIENT that's set in the backend configuration for the web access is different from the one needed to decode the id_token from the app. This is because of course apple does not allow a service ID to be the same as the bundle ID.

@sposs I solved a similar scenario extending the base AppleID in local for my project, something like this maybe could help you:

## Define new backends

The name will be used to fetch concrete settings and link it at the `provider` field (or the `/jwt-pair/$name` URI)

```python
# file: an_app/backends.py

from social_core.backends.apple import AppleIdAuth

"""
It provides 2 backends that extends the AppleID base auth

Useful to be able to define different configs depending on the singin scenario
"""
class AppleIdWebAuth(AppleIdAuth):
    """
    Default AppleID backend for app logins. Is the same than AppleIdAuth, but helps to ensure that provider name does not change.
    """

    name = "apple-id"


class AppleIdAppAuth(AppleIdAuth):
    """
    Default AppleID backend for app logins.  Will load SOCIAL_AUTH_APPLE_ID_FROM_APP_* settings and expose it as provider = "apple-id-from-app" (or your URL /jwt-pair/apple-id-from-app)
    """
    name = "apple-id-from-app"

```

## Load and apply different settings for each backend

```python
file: settings/__init__.py  # or your backends loader

...

AUTHENTICATION_BACKENDS = [
    ...
    # 'social_core.backends.apple.AppleIdAuth', 
    'users.backends.AppleIdAppAuth',
    'users.backends.AppleIdWebAuth',
    ...
]

SOCIAL_AUTH_APPLE_ID = 'com.client.web' 
SOCIAL_AUTH_APPLE_KEY = '...'
...

# FROM APP Settings
SOCIAL_AUTH_APPLE_ID_FROM_APP_CLIENT = 'com.client.app'
SOCIAL_AUTH_APPLE_ID_FROM_APP_KEY = '...'
...

```

https://gist.github.com/XaviTorello/e7793b1c6f68fe60a0036d581d3cd361

pizzapanther:
> I've been trying to use this to authenticate from a device: the iOS device obtains a id_token and sends it. Unfortunately, the verification fails always, and I had to put `verify=False` in the `jwt.decode` step. Any idea why the verification would fail?

I started having this problem too and it was working before.

XaviTorello:
@omab what do you think, it can be merged?

The related doc has been provided at https://github.com/python-social-auth/social-docs/pull/68

sposs:
Hi,
I actually still have a problem: I have realized I had not defined the scope. Once I do that, I keep getting the AuthStateMissing exception. I understand that it's because the 'state' parameter cannot be found in the session. My session is produced by the SessionMiddleware. From the doc, https://docs.djangoproject.com/en/3.0/ref/settings/#session-cookie-samesite, I understand that the default config Lax prevents the cookies from being sent when the request uses POST. But, that's what Apple requires to sent the name/email (response_mode=form_post). So the session key is not sent, and the state is not found and the request fails. Any one else had the issue? Any hints on how to address that?

sposs:
I can actually confirm that setting 
```python
SESSION_COOKIE_SAMESITE = None
```
solves the issue...

sposs:
Thanks!!

requirements-base.txt
```diff
@@ -3,3 +3,4 @@ oauthlib>=1.0.3
 requests-oauthlib>=0.6.1
 six>=1.10.0
 PyJWT>=1.4.0
+cryptography>=1.4
```
stianjensen, line 6: Maybe there's something I don't understand about what this package exports - but how is it used in this PR / why was it added?

grazor: This is required by PyJWT to use RSA encryption. https://pyjwt.readthedocs.io/en/latest/installation.html#cryptographic-dependencies-optional

knowmelife:
@grazor Thanks tons for your work on this; This example was so necessary for my project, I am glad I found out about it after this was merged.

 I had some questions about how you tested with the IOS SDK - if so, how was it called without storing the client_secret/client_id on the IOS device. 

Which endpoint was hit?
What were the parameters of the API call? 
Did the parameters still include the client_id/client_secret? If so, what is the purpose of code_authorization flow?  Wouldn't it be just as insecure as as implicit flow because the client_id/client_secret could be decoded from the App store release.

Anyways, I appreciate your contributions and am excited hear back form you.
Spencer

eoinnoble:
@knowmelife – I can't speak to the purpose of the auth flows, but I was able to get the iOS handshake to work in my Django app by creating a new endpoint that accepted authorization code that came back from Apple and passing that to the `complete` action. Do remember that the iOS app will likely generate that code using a different client ID, so make sure you're initializing your Apple backend with the correct one.

@sposs – changing `SESSION_COOKIE_SAMESITE` seemed like a bit of a sledgehammer approach to me, so I ended up creating a small piece of middleware that would take an Apple complete request and redirect that to my endpoint with `request.POST.urlencode()` appended to the redirect URL. That allowed Django to use the session cookie.

jonesbasf:
@eoinnoble - would you mind sharing what you have done. I am having a terrible time getting my backend to properly process the token from the iOS app.

eoinnoble:
> @eoinnoble - would you mind sharing what you have done. I am having a terrible time getting my backend to properly process the token from the iOS app.

Can you be more specific about the problems you're seeing @jonesbasf? I have created a very basic [gist](https://gist.github.com/eoinnoble/8494522161f043762237260786433c33) that might be helpful, but without knowing more about what you're running into it's hard to give you advice.

jonesbasf:
> > @eoinnoble - would you mind sharing what you have done. I am having a terrible time getting my backend to properly process the token from the iOS app.
> 
> Can you be more specific about the problems you're seeing @jonesbasf? I have created a very basic [gist](https://gist.github.com/eoinnoble/8494522161f043762237260786433c33) that might be helpful, but without knowing more about what you're running into it's hard to give you advice.

I actually figured out what my issue was. I had been following a writeup from someone and they had instructed to send the wrong thing to the backend. After more research, I determined that I had to send the token from the ASAuthorizationAppleIDCredential into the backend and everything started working. 

Just incase anyone else runs into this, here is what I did:

    NSData *authCode = appleCredentials.identityToken;
    NSString *token = [[NSString alloc] initWithData:authCode encoding:NSUTF8StringEncoding];
    
    NSURL * serviceURL = [[SNFConfig sharedInstance] apiURLForPath:@"token_auth/apple-id"];
    NSURLRequest * request = [NSURLRequest formURLEncodedPostRequestWithURL:serviceURL variables:@{@"access_token":token}];

appleCredentials is a ASAuthorizationAppleIDCredential object

PR #429, opened 2020-02-05
Fix typos.

asherf:


social_core/tests/backends/test_nationbuilder.py
```diff
@@ -133,7 +133,7 @@ class NationBuilderOAuth2Test(OAuth2Test):
             'twitter_friends_count': None,
             'facebook_username': None,
             'active_customer_started_at': None,
-            'pf_strat_id': None,
```
omab: This is an actual field for this backend: https://nationbuilder.com/fields_available_for_import

omab: Thanks

PR #428, opened 2020-02-05
[Github] Pass access token in a header.

asherf:
https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters
> GitHub is deprecating authentication to the GitHub API using query parameters, such as using a access_token query parameter .......

https://developer.github.com/v3/auth/#basic-authentication

PR #431, opened 2020-02-06
GitHub access token as basic auth.

PPingot:
GitHub access token as parameter is deprecated,
basic auth is recommended instead.


timabbott:
Can merging this be prioritized?  Today, GitHub started sending emails to any software using the old API every 3 days, which is pretty annoying and also suggests they may drop support for the interface imminently.  Thanks!

PPingot:
There is a better PR for this #428. I noticed it after sending this one. #428 is simpler, does not require base64 at all.

PR #432, opened 2020-02-07
Fix user_details in user pipeline to allow model attributes to be updated

rbricheno:
This has previously been discussed extensively in [issue #263](https://github.com/python-social-auth/social-core/issues/263) and there was a candidate [pull request #348](https://github.com/python-social-auth/social-core/pull/348) to fix it, but that currently has a failing test and hasn't been updated for a while. I made basically the same pull request but fixed the test.

I consider the current behaviour (i.e. not updating any field with new information from the social auth provider) to be a bug. It was [originally added in 2017](https://github.com/python-social-auth/social-core/pull/110).

The functionality requested in that PR is already provided (through configuration of `SOCIAL_AUTH_PROTECTED_USER_FIELDS`) since 2014 when it was explicitly added in commit  [0e411aa2a78136c99ffd36fde232ceef7eb252e3](https://github.com/omab/python-social-auth/commit/0e411aa2a78136c99ffd36fde232ceef7eb252e3#diff-809e16b5b0c2142b0a77fccfeec91e4a)

The 2017 change modified social-core to *never* update any attribute from a social auth provider if it is already set in the model.

The original (pre-PR#110) logic:
```
                current_value = getattr(user, name, None)
                if not current_value or name not in protected:
                    changed |= current_value != value
                    setattr(user, name, value)
```
appears to have been intended to always update the value if the current value is not set. That's understandable; if the value of an attribute wasn't set, and then we are given a value, we always want to set the attribute.

The logic which replaced it was:
```
        current_value = getattr(user, name, None)
        if current_value:
            continue

        if name in protected:
            continue

        if current_value == value:
            continue

        changed = True
        setattr(user, name, value)
```
This `continue`s if current value is `True`-ish. That's wrong! It has the effect of *never* setting an attribute if we have ever set it in the past, so we do not receive updates when the social auth provider tells us new information.

For example, my "display name" is always wrong in my app which uses an authentication provider via social-core, even if I update my display name in the authentication provider.

Currently, the code looks like this:
```
        current_value = getattr(user, name, None)
        if current_value or current_value == value:
            continue

        changed = True
        setattr(user, name, value)
```
which preserves the newer incorrect logic. The fix is to simply compare the values, and only continue if they are the same, ignoring whether or not there is a current value:
```
        if current_value == value:
            continue
```

Fixing this also broke a test. The failure was:
```
social_core/tests/backends/test_universe.py::UniverseAuth2Test::test_partial_pipeline FAILED
...
=============================================== FAILURES ===============================================
_______________________________ UniverseAuth2Test.test_partial_pipeline ________________________________

self = <social_core.tests.backends.test_universe.UniverseAuth2Test testMethod=test_partial_pipeline>

    def test_partial_pipeline(self):
>       self.do_partial_pipeline()

social_core/tests/backends/test_universe.py:32: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
social_core/tests/backends/base.py:150: in do_partial_pipeline
    self.assertEqual(user.slug, slug)
E   AssertionError: 'scott-vitale' != 'foo-bar'
E   - scott-vitale
E   + foo-bar
```
It seemed to make sense to fix this in `tests/backends/test_universe.py` as that is currently the only test which involves a slug, although it could equally have been fixed by modifying `pipeline_slug_handling` in `tests/backends/base.py`.

Finally, this PR also removes a related comment (which has been wrong since the PR which introduced this bug was merged) as per the [suggestion by SpoonMeiser](https://github.com/python-social-auth/social-core/pull/348#issuecomment-510900654).

Closes https://github.com/python-social-auth/social-core/issues/263


Binnette:
Hi @rbricheno , thanks for your PR. Hope this gonna be merge and release soon.
Have a nice day.

dadon-david:
Would love to see this PR Merged !!

LiamK:
Thank you, to the people who worked on this!  Hope it can be merged soon :-)

Issue #437, opened 2020-02-18
The Kakao Talk endpoint has been changed.


elastic7327:
The version of the Kakao Talk endpoint has been changed.

AS-IS:  `/v1/user/me` 

TO-BE:  `/v2/user/me`

rbricheno:
As per [the Korean documentation](https://translate.google.com/translate?hl=en&sl=ko&u=https://developers.kakao.com/docs/restapi/quick-reference&prev=search) there is another change required (as well as just moving the URL).

The returned object now has a different structure, and fields which were previously named `kaccount_...` are now all found within the new `kakao_account` sub-object.

Now, to get the email, where previously we said:

`kaccount_email = response.get('kaccount_email', '')`

we now have to say:

```
kakao_account = response.get('kakao_account', '')
kaccount_email = kakao_account.get('email', '')
```

(this information hasn't made it into the English docs yet).

This change is included in PR #441 from @KiyeopYang

PR #438, opened 2020-02-18
[feature/#437/kakao_backend_version_change] kakao route version chang…

elastic7327:
…e v1 -> v2

https://github.com/python-social-auth/social-core/issues/437


omab:
Fixed by #441

Issue #439, opened 2020-02-19
user.is_authenticated() as a method is removed in Django 2


mraarif:
as per Django 2.0 [release notes ](https://docs.djangoproject.com/en/3.0/releases/2.0/), `user.is_authenticated()` is removed, it was deprecated in Django 1.10.
will follow up with a PR to remove its usage from `scoial_core/utils.py`

omab:
This is the core library, not the Django extension.

PR #440, opened 2020-02-19
Removed the deprecated usage of user.is_authenticated()

mraarif:
removed the usage of `user.is_authenticated()` to replace it to use a property instead to fix 

> RemovedInDjango20Warning

details in issue #439

### DIFF
--- a/social_core/utils.py
+++ b/social_core/utils.py
@@ -112,7 +112,7 @@ def sanitize_redirect(hosts, redirect_to):
 def user_is_authenticated(user):
     if user and hasattr(user, 'is_authenticated'):
         if isinstance(user.is_authenticated, collections.Callable):
-            authenticated = user.is_authenticated()
+            authenticated = user.is_authenticated
         else:
             authenticated = user.is_authenticated
     elif user:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Use `unidecode` to cleanup usernames from unicode characters
 - Update Twitch API support from v3 to v5
 - Properly setup `pytest` version for Python2 and Python3
+- Removed usage of `user.is_authenticated()`
 
 ## [3.2.0](https://github.com/python-social-auth/social-core/releases/tag/3.2.0) - 2019-05-30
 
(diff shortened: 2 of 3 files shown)


social_core/utils.py
```diff
@@ -112,7 +112,7 @@ def sanitize_redirect(hosts, redirect_to):
 def user_is_authenticated(user):
     if user and hasattr(user, 'is_authenticated'):
         if isinstance(user.is_authenticated, collections.Callable):
-            authenticated = user.is_authenticated()
+            authenticated = user.is_authenticated
         else:
             authenticated = user.is_authenticated
```
atodorov: Your change seems to make the entire if condition unnecessary b/c both branches become the same.

social_core/utils.py
```diff
@@ -112,7 +112,7 @@ def sanitize_redirect(hosts, redirect_to):
 def user_is_authenticated(user):
     if user and hasattr(user, 'is_authenticated'):
         if isinstance(user.is_authenticated, collections.Callable):
-            authenticated = user.is_authenticated()
```
omab: This is the `core` library, and as such, it expects the `is_authenticated` to exist, but takes into consideration the option of it being a callable, that's the reason for the `if` before it.

atodorov: @omab are you saying that changes in this code block should be reverted ? That is how I understand it, although I don't understand when `is_authenticated` can be a callable. Maybe that applies to other frameworks, and not Django. 

If this is the case then this entire PR should not be needed. Instead just make sure tests cover the 2 branches of the if statement.

social_core/utils.py
mraarif: as of Django 1.10 [release](https://docs.djangoproject.com/en/3.0/releases/1.10/#using-user-is-authenticated-and-user-is-anonymous-as-methods) `is_authenticated` as a method was deprecated and introduced as a property instead and was removed in Django 2.0 as mentioned in issue #439 

to get rid of the deprecation warning and remove its usage as a method, I think we can safely remove this `if else` condition.

social_core/utils.py
omab: @atodorov, that's what I'm saying, not reverted, just that this PR is not needed, `is_authenticated()` was deprecated on Django, but this is the core library and needs to take into consideration other frameworks or custom implementations, on Flask is really easy to define your own user model. Ideally, this code should be in the corresponding strategies, have a `strategy.is_authenticated(user)` there that wraps the framework implementation.

nijel:
Closing as I agree with @omab here - such change would break usages outside Django. The other reason is that Django removed the depreciation warning meanwhile and `is_authenticated` is no longer callable there. The code works just fine now and is compatible with both ways to implement `is_authenticated`.

PR #441, opened 2020-02-20
[URGENT] kakao v1 -> v2

KiyeopYang:
Fixes #437

PR #443, opened 2020-02-27
change Kakao Version v1 => v2

wkddnjset:
Kakao no longer supports v1.
so if use v1, can causes an unknown error.


omab:
Fixed by #441

PR #452, opened 2020-04-16
Use passed in access_token rather than the response object - Fix #450

### LINKED ISSUE #450: Azure AD Tenant throws exception "AttributeError: 'NoneType' object has no attribute 'get'"
Attempting to verify Azure AD tenant token

### Expected behaviour

Token verification should succeed

### Actual behaviour

Exception thrown:

AttributeError: 'NoneType' object has no attribute 'get'

### Any other comments?

Comparing the azuread_tenant.py function `user_data` and the azuread.py `user_data` function reveals significant differences.  Specifically, `user_data` in azured.py checks for `response` being `None`.

Using the standard azuread backend in place of the tenant backend is a work-around since there are no tenant-specific keys for token validation

paulw11:
## Proposed changes

Fixes an error "'NoneType' object has no attribute 'get'" when using the azure_tenant backend to validate a JWT.  Fix #450

## Types of changes

- [X ] Bugfix (non-breaking change which fixes an issue)

## Checklist

your code._

## Other information

### DIFF
--- a/social_core/backends/azuread_tenant.py
+++ b/social_core/backends/azuread_tenant.py
@@ -94,7 +94,7 @@ class AzureADTenantOAuth2(AzureADOAuth2):
 
     def user_data(self, access_token, *args, **kwargs):
         response = kwargs.get('response')
-        id_token = response.get('id_token')
+        id_token = access_token
 
         # decode the JWT header as JSON dict
         jwt_header = json.loads(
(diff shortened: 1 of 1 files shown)


dacevedo12:
Thanks! I just had this exact problem. Any change of getting this merged soon? I'll patch it meanwhile

omab:
PR merged manually to add the corresponding changelog entry, but Github fails to pick up the change. Thanks!

dacevedo12:
Oh no, I just realized the reason why azure.py uses that condition. If you auth programatically using an oauth2 token response is None and access token is your JWT. But if you auth through the view then response exists and access token is NOT a JWT.

Welp, this PR fixed one thing while breaking another

paulw11:
So, we need something like:
```Python
 def user_data(self, access_token, *args, **kwargs):
      
    response = kwargs.get('response')
    response_token = response.get('id_token')
    id_token = response_token if response_token else  access_token
         
```

Issue #419, opened 2020-01-08
Instagram closed new app registration. Provider should be refactored or should be deleted


yurabysaha:
According to the Instagram announcement:

> UPDATE: Starting October 15, 2019, new client registration and permission review on Instagram API platform are discontinued in favor of the Instagram Basic Display API. 

Need to refactor instagram backend provider, because now its not a working. 

instagram backend provider use `https://api.instagram.com/v1/users/self` for get user data, but now it should be `https://graph.instagram.com/me`. [link to the doc](https://developers.facebook.com/docs/instagram-basic-display-api/reference/me)

 Also need refactor `get_user_details` and `get_user_id` methods because response is also changed now.

But on [Instagram basic display api page](https://developers.facebook.com/docs/instagram-basic-display-api/)  we can see announcement in Limitations block.
> Instagram Basic Display is not an authentication solution. Data returned by the API cannot be used to authenticate your app users or log them into your app. If you need an authentication solution we recommend using Facebook Login instead

So maybe should just remove this provider from backend.

KimVuu:
Is there a countermeasure for this?

EralpB:
any news about this? how are you people solving this?

yurabysaha:
Just remove instagram social and use facebook

Nadjedu:
Is there a workaround or a fix planned?

PR #460, opened 2020-05-07
Instagram backend refactor

Nadjedu:
## Proposed changes

Instagram has deprecated most of the endpoints on api.instagram and moved all data access to graph.instagram. 
**Short story**: Instagram Basic Display is no longer an authentication solution. 
**Long story**: Instagram Basic Display only works on accounts that are not creator or business accounts and the endpoint will not return data (email, etc) that can be used to authenticate users. However, for creator and business accounts, the facebook backend is the authentication solution. I propose that we keep the Instagram backend for users who want to associate Instagram personal (non-creator and non-business) accounts with authenticated users. It is important to note that Facebook does not verify apps (social_core library users) that use this as an authentication method.

## Types of changes

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Updated list of default user protected fields to include admin flags and password
+- Instagram: Update to use the latest Graph API v3.2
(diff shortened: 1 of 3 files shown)


Nadjedu:
This pull request is a response to the open issue: #419

Issue #464, opened 2020-05-26
[Apple ID] Logging in with iOS and a web-based service + bug with User updating


maiiku:
### Expected behaviour

* I should be bale to use Apple ID backend to login via both iOS and web-form/other approach.
* Consecutive logins should not update User firs/last name with empty strings 

### Actual behaviour

* I can now specify either iOS app to login or service-based up to login sine the audience passed to JWT token decoding is set to the CLIENT setting.
* On consecutive logins if pipeline uses `social_core.pipeline.user.user_details` user first nad last name is overriden with empty strings

### What are the steps to reproduce this issue?

* For name overriding log in a second time (apple only supplies name on first login when account is paired with given Apple ID) 

* For signing in with both iOS and other services:
1. configure your App to use Apple ID, set it it as `primary App id` for Apple ID on Apple Developer portal
2. configure service with proper domain / return urls on Apple Developer portal
3. set up Apple backend as per documentation fro social-auth
4. set CLIENT setting to bundle ID
5. try to logging in with iOS (success)
6. try logging in via web form (failure)
7. change CLIENT to service ID
8. try to logging in with iOS (failure)
9. try logging in via web form (success)

PR #465, opened 2020-05-26
Some fixes for Apple backend:

maiiku:
## Proposed changes

- add ability for users to supply aud for JWT as a list. This enables apple backednd to be used by both iOS login and web from / other sources login, because iOS provides bundle id as client and web from login requires service id to be supplied (to match against defined return urls and allowed domains)
- if there's no first/last name in the response return None so that the User model is not updated to empty strings if user_details is used in the pipeline

## Types of changes

## Checklist

your code._

solves https://github.com/python-social-auth/social-core/issues/464


omab:
Thanks!

Issue #474, opened 2020-06-12
Send proxies while making requests to authorization endpoints


bajibabu9:
Hi, I was trying to setup a django app and I'm using python social auth to sign in with Google. The django app is running on a Linux machine where we don't have direct access to external websites, should use proxy to make requests to any websites like google.com, linkedin.com, etc.,

So, the problem is during google oauth flow, it's unable to make requests to https://accounts.google.com the connection is timing out after a couple of mins due to access to other websites blocked on the machine

Would like to support sending proxies in the following function
https://github.com/python-social-auth/social-core/blob/master/social_core/backends/base.py#L216

```python
if self.setting('PROXIES'):
     kwargs.setdefault('proxies', self.setting('PROXIES'))
```
We can set the variable in django settings like this
SOCIAL_AUTH_PROXIES = {'http:' 'http://IP: PORT', 'https': 'https://IP: PORT'}

bajibabu9:
Closing this as PR is merged.

PR #475, opened 2020-06-12
Send proxies in request

bajibabu9:
Connected to https://github.com/python-social-auth/social-core/issues/474


bajibabu9:
Thank you @omab

PR #481, opened 2020-06-18
Make WHITELIST_* settings properly case insensitive

danlamanna:
## Proposed changes

With a setting of `SOCIAL_AUTH_WHITELISTED_DOMAINS` set to `gmail.com`, a separate system storing
the domain as `Gmail.com` would be rejected. According to [RFC 1035](https://tools.ietf.org/html/rfc1035) the domain should be checked in a case insensitive manner.

This applies only to the domain portion of whitelisted email addresses as well, e.g. `foo@bar.com` is distinct from `FOO@bar.com`.

I was unable to make the change as cleanly as I would have liked, since it appears that misconfiguring the `SOCIAL_AUTH_WHITELISTED_EMAILS` setting by giving it a domain instead of a valid email is implicitly supported, see https://github.com/python-social-auth/social-core/blob/dddb1784fa54468a09f552ca299f60017f4f6aa9/social_core/tests/backends/test_dummy.py#L111.

## Types of changes

## Checklist

	I was unable to get everything passing locally before I made the change. But no additional tests fail.

### DIFF
--- a/social_core/backends/base.py
+++ b/social_core/backends/base.py
@@ -148,12 +148,24 @@ class BaseAuth(object):
     def auth_allowed(self, response, details):
         """Return True if the user should be allowed to authenticate, by
         default check if email is whitelisted (if there's a whitelist)"""
-        emails = self.setting('WHITELISTED_EMAILS', [])
-        domains = self.setting('WHITELISTED_DOMAINS', [])
-        email = details.get('email')
+        def lower_email_domain(email):
+            """Lowercase an email safely by leaving the local-part alone and
+            only lowercasing the domain."""
+            if email is None:
+                return
+
+            try:
+                local_part, domain = email.split('@', 1)
+            except ValueError:
+                return email  # the email is malformed, pass it through unchanged
+            return '@'.join([local_part, domain.lower()])
+
+        emails = [lower_email_domain(email) for email in self.setting('WHITELISTED_EMAILS', [])]
+        domains = [domain.lower() for domain in self.setting('WHITELISTED_DOMAINS', [])]
+        email = lower_email_domain(details.get('email'))
         allowed = True
         if email and (emails or domains):
-            domain = email.split('@', 1)[1]
+            domain = email.split('@', 1)[1].lower()
             allowed = email in emails or domain in domains
         return allowed
 
--- a/social_core/backends/base.py
+++ b/social_core/backends/base.py
@@ -150,9 +150,10 @@ class BaseAuth(object):
         default check if email is whitelisted (if there's a whitelist)"""
         emails = [email.lower() for email in self.setting('WHITELISTED_EMAILS', [])]
         domains = [domain.lower() for domain in self.setting('WHITELISTED_DOMAINS', [])]
+        email = details.get('email')
         allowed = True
         if email and (emails or domains):
-            email = details.get('email').lower()
+            email = email.lower()
             domain = email.split('@', 1)[1]
             allowed = email in emails or domain in domains
         return allowed
--- a/social_core/backends/base.py
+++ b/social_core/backends/base.py
@@ -148,24 +148,12 @@ class BaseAuth(object):
     def auth_allowed(self, response, details):
         """Return True if the user should be allowed to authenticate, by
         default check if email is whitelisted (if there's a whitelist)"""
-        def lower_email_domain(email):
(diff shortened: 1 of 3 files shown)


danlamanna:
This isn't stale, it still needs review.

social_core/backends/base.py
```diff
@@ -148,12 +148,24 @@ def extra_data(self, user, uid, response, details=None, *args, **kwargs):
...
+            """Lowercase an email safely by leaving the local-part alone and
+            only lowercasing the domain."""
+            if email is None:
+                return
+
+            try:
+                local_part, domain = email.split('@', 1)
+            except ValueError:
+                return email  # the email is malformed, pass it through unchanged
+            return '@'.join([local_part, domain.lower()])
```
nijel, line 161: I'd assume that whole email is case-insentivite. That would make the whole code a lot simpler and matches reality, see https://stackoverflow.com/a/9808332/225718

social_core/backends/base.py
```diff
@@ -148,12 +148,24 @@ def extra_data(self, user, uid, response, details=None, *args, **kwargs):
...
+
+        emails = [lower_email_domain(email) for email in self.setting('WHITELISTED_EMAILS', [])]
+        domains = [domain.lower() for domain in self.setting('WHITELISTED_DOMAINS', [])]
+        email = lower_email_domain(details.get('email'))
```
nijel, line 165:
```suggestion
        emails = [email.lower() for email in self.setting('WHITELISTED_EMAILS', [])]
        domains = [domain.lower() for domain in self.setting('WHITELISTED_DOMAINS', [])]
```

* Remove lower_email_domain code, it is IMHO not necessary
* Remove lowercasing email here as it might be None as well

social_core/backends/base.py
```diff
@@ -148,12 +148,24 @@ def extra_data(self, user, uid, response, details=None, *args, **kwargs):
...
         allowed = True
         if email and (emails or domains):
-            domain = email.split('@', 1)[1]
+            domain = email.split('@', 1)[1].lower()
```
nijel, line 168:
```suggestion
            email = details.get('email').lower()
            domain = email.split('@', 1)[1]
```

nijel (changes requested):

social_core/backends/base.py
```diff
@@ -148,11 +148,11 @@ def extra_data(self, user, uid, response, details=None, *args, **kwargs):
...
         """Return True if the user should be allowed to authenticate, by
         default check if email is whitelisted (if there's a whitelist)"""
-        emails = self.setting('WHITELISTED_EMAILS', [])
-        domains = self.setting('WHITELISTED_DOMAINS', [])
-        email = details.get('email')
+        emails = [email.lower() for email in self.setting('WHITELISTED_EMAILS', [])]
+        domains = [domain.lower() for domain in self.setting('WHITELISTED_DOMAINS', [])]
         allowed = True
         if email and (emails or domains):
+            email = details.get('email').lower()
```
nijel, line 155:
```suggestion
        email = details.get('email')
        allowed = True
        if email and (emails or domains):
            email = email.lower()
```

social_core/tests/backends/test_dummy.py
```diff
@@ -98,6 +98,18 @@ def test_invalid_login(self):
         with self.assertRaises(AuthForbidden):
             self.do_login()
 
+    def test_invalid_login_case_sensitive_local_part(self):
+        self.strategy.set_settings({
+            'SOCIAL_AUTH_WHITELISTED_EMAILS': ['fOo@bar.com']
+        })
+        with self.assertRaises(AuthForbidden):
+            self.do_login()
```
nijel, line 106:
```suggestion
    def test_login_case_sensitive_local_part(self):
        self.strategy.set_settings({
            'SOCIAL_AUTH_WHITELISTED_EMAILS': ['fOo@bar.com']
        })
        self.do_login()
```

social_core/tests/backends/test_dummy.py
```diff
@@ -98,6 +98,18 @@ def test_invalid_login(self):
...
+
+    def test_invalid_login_case_sensitive_domain(self):
```
nijel, line 108:
```suggestion
    def test_login_case_sensitive_domain(self):
```

nijel:

PR #490, opened 2020-07-07
Fix apple and saml backend filling user_details with `None` values.

chdinesh1089:
## Proposed changes
Fixes apple and saml backends filling user_details with values containing `None`. This project [generally uses `''` instead of `None`](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/base.py#L176). Most other backends use `''`.

## Types of changes

## Checklist

your code._

## Other information

### DIFF
--- a/social_core/backends/apple.py
+++ b/social_core/backends/apple.py
@@ -142,10 +142,9 @@ class AppleIdAuth(BaseOAuth2):
 
         email = response.get('email', '')
         apple_id = response.get(self.ID_KEY, '')
-        # prevent updating User with empty strings
         user_details = {
-            'first_name': first_name or None,
-            'last_name': last_name or None,
+            'first_name': first_name,
+            'last_name': last_name,
             'email': email,
         }
         if email and self.setting('EMAIL_AS_USERNAME'):
--- a/social_core/backends/saml.py
+++ b/social_core/backends/saml.py
@@ -79,7 +79,7 @@ class SAMLIdentityProvider(object):
             if len(value):
                 value = value[0]
             else:
-                value = None
+                value = ''
         return value
 
     @property


chdinesh1089:
@maiiku This PR reverts your change in #465 to have `None` for firstName and lastName in apple backend when it does not get name details. This could break things in your project/s if it depends on `firstName` and `lastName` to be `None`.

maiiku:
@chdinesh1089 this breaks things in another way, that is if your project has those details this will override First and Last Names to empty strings on next login. This is why I changed it to `None` in the first place, as `None` is omitted in update process by social-auth. If you wish to do it this way you should also modify modify relevant code for user updating to ignore empty strings. 

However I think it is your project that should handle `None` values for first nad last name in the first place, as this is dependent on your project nad social-auth, and that's why I proposed previous solution as opposed to changing user updating process in social-auth.

chdinesh1089:
@maiiku Since most other backends in `social-core` give empty strings for such cases, the projects supporting multiple authentication methods dependent on `social-core` will have code to handle empty strings and not `None`. Making it `None` would require everyone to handle the possibility of `None` just for apple and empty strings for all other backends. This inconsistency is not great.

Also, it is mentioned in this project's code to avoid `None` values. https://github.com/python-social-auth/social-core/blob/master/social_core/backends/base.py#L176

mateuszmandera:
I think @maiiku is right here, looking at this code from `social_core.pipeline.user.user_details`:
```
    field_mapping = strategy.setting('USER_FIELD_MAPPING', {}, backend)
    for name, value in details.items():
        # Convert to existing user field if mapping exists
        name = field_mapping.get(name, name)
        if value is None or not hasattr(user, name) or name in protected:
            continue
```

it seems that indeed the intent is that if the field (so first_name and last_name in our apple case) should be considered as "no value sent" and shouldn't overwrite the existing data on the user, then the value should be set to `None`. So it seems right to have the `None`s in the Apple backend.

The question would be what should be the right behavior for SAML.

maiiku:
@chdinesh1089 then as I said, if you wish to change this you should also propose a change to `user_details` in `social_core/pipeline/user.py` which currently only skips None values:
```python
    for name, value in details.items():
        # Convert to existing user field if mapping exists
        name = field_mapping.get(name, name)
        if value is None or not hasattr(user, name) or name in protected:
            continue

        current_value = getattr(user, name, None)
        if current_value == value:
            continue

        changed = True
        setattr(user, name, value)
```
as this actively overwrites in your database user details. 
Perhaps other backends do not return Nones on a second login and apple does here and hence the problem is not visible.

`get_user_names` you mentioned was specifically coded to handle `None` values, not as a warning against them

mateuszmandera:
It seems to me that `user_details` should not overwrite user attributes with empty strings, so that might be right change to make. But I don't know if the current behavior isn't being relied on by other projects/backends, so I don't feel confident in this opinion.

PR #515, opened 2020-10-24
Remove discontinued GoogleOpenId backend

andersk:
## Proposed changes

The Google OpenID backend was discontinued on the server side in 2015, in favor of OpenID Connect (our `GoogleOpenIdConnect` backend).

https://support.google.com/accounts/answer/6206245

Fixes #462. Fixes #472.

This is needed to get Travis passing again.

## Types of changes

Please check the type of change your PR introduces:

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (PEP8, lint, formatting, renaming, etc)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build related changes (build process, tests runner, etc)
- [ ] Other (please describe):

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask. We're here to
help! This is simply a reminder of what we are going to look for before merging
your code._

- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works

## Other information

Any other information that is important to this PR such as screenshots of how
the component looks before and after the change.

Issue #472, opened 2020-06-04
The url used by the GoogleOpenId backend seems to be deprecated


JoseKilo:
The url used by the `GoogleOpenId` backend seems to be deprecated:
https://github.com/python-social-auth/social-core/blob/master/social_core/backends/google.py#L176

If you try to access https://www.google.com/accounts/o8/id a 404 is returned.

### Expected behaviour

Probably `GoogleOpenId` needs to be removed and the docs updated.

### Actual behaviour

An error is returned when this backend is used.

### What are the steps to reproduce this issue?

1. Configure a project with:
```
AUTHENTICATION_BACKENDS = (
    ...
    'social_core.backends.google.GoogleOpenId',
    ...
)
```
2. Try to access http://127.0.0.1:8000/auth/login/google/
3. A 500 response is returned.

### Any logs, error output, etc?

The stack trace of the error:

```
Traceback (most recent call last):
  File "/home/jose/workspace/project/.venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/home/jose/workspace/project/.venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/jose/workspace/project/.venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/jose/workspace/project/.venv/lib/python3.8/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/home/jose/workspace/project/.venv/lib/python3.8/site-packages/social_django/utils.py", line 49, in wrapper
    return func(request, backend, *args, **kwargs)
  File "/home/jose/workspace/project/.venv/lib/python3.8/site-packages/social_django/views.py", line 23, in auth
    return do_auth(request.backend, redirect_name=REDIRECT_FIELD_NAME)
  File "/home/jose/workspace/project/.venv/lib/python3.8/site-packages/social_core/actions.py", line 29, in do_auth
    return backend.start()
  File "/home/jose/workspace/project/.venv/lib/python3.8/site-packages/social_core/backends/base.py", line 34, in start
    if self.uses_redirect():
[...]

Issue #462, opened 2020-05-19
Code should be mocked properly for test GoogleOpenIdTest::test_login


tkdchen:
### Expected behaviour

Relative code should be mocked properly for test `GoogleOpenIdTest::test_login` in order to not visit the real Google OpenId URL https://www.google.com/accounts/o8/id.

### Actual behaviour

When running test `GoogleOpenIdTest::test_login`, in method `openid_request`, `self.consumer().begin(...)` will visit the real URL. That could cause the test hangs when no network or the Google is just not accessible at the moment.

### What are the steps to reproduce this issue?

Input clear steps to reproduce the issue for a maintainer.

1. clone social-core
2. create a virtualenv for the project
3. start to run tests within a testenv, in my case, it is `tox -e py36`

PR #531, opened 2020-12-01
[Okta] Properly join configuration URLs

### LINKED ISSUE #524: Okta OIDC backend fails token validation with custom auth servers
### Expected behaviour

Tokens issued by Okta's Org and Custom Authorization servers should pass validation.

For a single Okta tenant, these might be valid auth server urls:
- Org Auth Server: https://dev-000000.okta.com/
- Default Custom: https://dev-000000.okta.com/oauth2/default/
- Additional Custom: https://dev-000000.okta.com/oauth2/$serverId/
- Customer Domain: https://$customerDomain/oauth2/$serverId/

These should be usable as values for `SOCIAL_AUTH_OKTA_OPENIDCONNECT_API_URL`

Further descriptions are available here: https://developer.okta.com/docs/concepts/auth-servers/

### Actual behaviour

As called out in the docs and the comments in @selten's initial merge request (https://github.com/python-social-auth/social-core/pull/333#issuecomment-512555486), the only value that works is:
`SOCIAL_AUTH_OKTA_OPENIDCONNECT_API_URL="https://dev-000000.okta.com/oauth2"`

This seems to be due to how the OIDC discovery endpoint URL is constructed. (https://github.com/python-social-auth/social-core/blob/master/social_core/backends/okta.py#L27). oidc_config() returns the json from the following url:

    >>> api_url = 'https://dev-709546.okta.com/oauth2'
    >>> well_known = '/.well-known/openid-configuration'
    >>> urljoin(api_url, well_known)
    'https://dev-000000.okta.com/.well-known/openid-configuration'

This JSON contains the metadata for the Org Auth Server. 

If the above is executed with the Default Custom auth server url we see:

    >>> api_url = 'https://dev-000000.okta.com/oauth2/default/'
    >>> well_known = '/.well-known/openid-configuration'
    >>> urljoin(api_url, well_known)
    'https://dev-000000.okta.com/.well-known/openid-configuration'
This returns the discovery metadata url for the Org Auth Server, rather than the expected value:

    'https://dev-000000.okta.com/oauth2/default/.well-known/openid-configuration'

This seems to be caused by urllib.parse.urljoin() treating the well_known path as an absolute url due to it's leading `/` which strips the auth server information. When validation occurs, it will fail due to different authorization servers being used compared to the jwks keys retrieved for validation. The authorization and token URLs are built with relative paths (https://github.com/python-social-auth/social-core/blob/master/social_core/backends/okta.py#L16), while the jwks_uri is retrieved from the discovery JSON.

Removing the leading `/` from the well-known path causes the URLs to build as expected.

    >>> from six.moves.urllib.parse import urljoin
    >>> auth_servers = (
        'https://dev-000000.okta.com/oauth2/',
        'https://dev-000000.okta.com/',
        'https://dev-000000.okta.com/oauth2/default/',
        'https://dev-000000.okta.com/oauth2/custom/',
        'https://customer.domain.com/oauth2/custom/',
    )
    >>> well_known = '/.well-known/openid-configuration'
    >>> for server in auth_servers:
            print(urljoin(server, well_known))
        
    https://dev-000000.okta.com/.well-known/openid-configuration
    https://dev-000000.okta.com/.well-known/openid-configuration
    https://dev-000000.okta.com/.well-known/openid-configuration
    https://dev-000000.okta.com/.well-known/openid-configuration
    https://customer.domain.com/.well-known/openid-configuration
    >>> well_known = '.well-known/openid-configuration'
    >>> for server in auth_servers:
            print(urljoin(server, well_known))
        
    https://dev-000000.okta.com/oauth2/.well-known/openid-configuration
    https://dev-000000.okta.com/.well-known/openid-configuration
    https://dev-000000.okta.com/oauth2/default/.well-known/openid-configuration
    https://dev-000000.okta.com/oauth2/custom/.well-known/openid-configuration
    https://customer.domain.com/oauth2/custom/.well-known/openid-configuration

Unfortunately, this isn't as straightforward as removing a single character, as it will break backwards compatibility. The API_URL suggested by the documentation isn't a valid one; it just happened to work with this bug. `https://dev-000000.okta.com/oauth2/.well-known/openid-configuration` will http/404.

stuharvey:
Closes #524.

## Proposed changes

This change ensures that the configuration endpoint `/.well-known/openid-configuration` is appended to the configured Okta API_URL rather than replacing the whole slug.

The current approach used to generate the URL Okta applications truncates the provided `API_URL` then appends `/.well-known/openid-configuration` to it. So for custom applications, `mycompany.okta.com/oauth2/<app-identifier>/.well-known/openid-configuration` is instead truncated to `mycompany.okta.com/.well-known/openid-configuration` which won't 404 but may provide incorrect information.

## Types of changes

## Checklist

Tests pass, but I couldn't find any lint config.

### DIFF
--- a/social_core/backends/okta.py
+++ b/social_core/backends/okta.py
@@ -24,7 +24,7 @@ class OktaMixin(object):
     def oidc_config(self):
         return self.get_json(
             self._url(
-                '/.well-known/openid-configuration?client_id={}'.format(
+                './.well-known/openid-configuration?client_id={}'.format(
                     self.setting('KEY')
                 )
             )


stuharvey:
Hi @omab, is there a way to rerun CI without pushing a new commit? It appears that the failure is transient, and tests pass locally.

yuekui:
> Hi @omab, is there a way to rerun CI without pushing a new commit? It appears that the failure is transient, and tests pass locally.

Hi @stuharvey you may trigger CI by pushing an empty commit like this:
```git commit --allow-empty -m "Trigger Build"```

But for the failure of this PR(in fact for all PRs of this project right now), re-run won't change the result, it's caused by GoogleOpenId test case and has been resolved in #515 , see also #462 #472

digismack:
> @stu-harvey So I'm having issues now because my okta configuration is expecting example.com/.well-known but all other URLS expect example.com/oauth2/v1/ So with this code change I get example.com/oauth2/.well-known. If I change the API_URL and remove /oauth2/ I get issues with the rest of my URLs. Any ideas?

We are also experiencing this issue. So far the fix seems to be ensuring that the configuration always includes a default or custom authorization server in the API_URL, but I realize that's not always possible.

I'm not sure that I would classify this library as not actively maintained. It's gotten a little better recently since they added another maintainer to the mix. I'm sure they would welcome a pull request to resolve the issue.

digismack:
> @stu-harvey So I'm having issues now because my okta configuration is expecting example.com/.well-known but all other URLS expect example.com/oauth2/v1/ So with this code change I get example.com/oauth2/.well-known. If I change the API_URL and remove /oauth2/ I get issues with the rest of my URLs. Any ideas?

We are also experiencing this issue. So far the fix seems to be ensuring that the configuration always includes a default or custom authorization server in the API_URL, but I realize that's not always possible.

I'm not sure that I would classify this library as not actively maintained. It's gotten a little better recently since they added another maintainer to the mix. I'm sure they would welcome a pull request to resolve the issue.

PR #554, opened 2021-01-28
Amends to social_core.backends.orcid.py

asencis:
## Proposed changes

Added a new USER_DATA_ID attribute for the ORCIDOAuth2 and ORCIDMemberOAuth2 classes (and their associated sandbox modes).

Alterations made to self.user_data() call to pass the access_token as a Bearer token to that endpoint. Reference: [https://github.com/ORCID/ORCID-Source/blob/master/orcid-web/ORCID_AUTH_WITH_OPENID_CONNECT.md#other-endpoints] for more details.

Alterations made to the get_user_details() as the orcid parameter is now longer accessible at response['orcid'] instead it is at response['orcid-identifier']['path'].

## Types of changes

## Checklist

### DIFF
--- a/social_core/backends/orcid.py
+++ b/social_core/backends/orcid.py
@@ -26,14 +26,15 @@ class ORCIDOAuth2(BaseOAuth2):
 
     def get_user_details(self, response):
         """Return user details from ORCID account"""
-        orcid_identifier = response.get('orcid-identifier', None)
+        orcid_identifier = response.get('orcid-identifier')
 
         fullname = response.get('name', '')
 
-        first_name = last_name = email = ''
+        first_name = last_name = email = username = ''
 
-        person = response.get('person', None)
+        person = response.get('person')
 
+        # Although we're checking here, the response will always have the orcid-identifier key:
         if orcid_identifier:
             username = orcid_identifier['path']
 
--- a/social_core/backends/orcid.py
+++ b/social_core/backends/orcid.py
@@ -1,9 +1,8 @@
 """
-ORCID OAuth2 Application backend, docs at:
+    ORCID OAuth2 Application backend, docs at:
     https://python-social-auth.readthedocs.io/en/latest/backends/orcid.html
 """
 from .oauth import BaseOAuth2
-
 
 class ORCIDOAuth2(BaseOAuth2):
     """ORCID OAuth2 authentication backend"""
@@ -11,6 +10,7 @@ class ORCIDOAuth2(BaseOAuth2):
     ID_KEY = 'orcid'
     AUTHORIZATION_URL = 'https://orcid.org/oauth/authorize'
     ACCESS_TOKEN_URL = 'https://orcid.org/oauth/token'
+    USER_ID_URL = 'https://orcid.org/oauth/userinfo'
     USER_DATA_URL = 'https://pub.orcid.org/v2.0/{}'
     DEFAULT_SCOPE = ['/authenticate']
     ACCESS_TOKEN_METHOD = 'POST'
@@ -26,9 +26,17 @@ class ORCIDOAuth2(BaseOAuth2):
 
     def get_user_details(self, response):
         """Return user details from ORCID account"""
+        orcid_identifier = response.get('orcid-identifier', None)
+
         fullname = response.get('name', '')
+
         first_name = last_name = email = ''
-        person = response.get('person')
+
+        person = response.get('person', None)
+
+        if orcid_identifier:
+            username = orcid_identifier['path']
+
         if person:
             name = person.get('name')
             if name:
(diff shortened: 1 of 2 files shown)


social_core/backends/orcid.py
```diff
@@ -26,9 +26,17 @@ def auth_params(self, state=None):
 
     def get_user_details(self, response):
         """Return user details from ORCID account"""
+        orcid_identifier = response.get('orcid-identifier', None)
```
nijel, line 29: The `None` parameter is not needed, it's the default value.

social_core/backends/orcid.py
```diff
@@ -26,9 +26,17 @@ def auth_params(self, state=None):
...
+
         fullname = response.get('name', '')
+
         first_name = last_name = email = ''
-        person = response.get('person')
+
+        person = response.get('person', None)
```
nijel, line 35: ...and same here.

social_core/backends/orcid.py
```diff
@@ -49,22 +57,48 @@ def get_user_details(self, response):
                     else:
                         email = emails_list[0].get('email', '')
 
-        return {'username': response.get('orcid'),
-                'email': email,
-                'fullname': fullname,
-                'first_name': first_name,
-                'last_name': last_name}
+        return {
+            'username': username,
```
nijel, line 61: The `username` parameter might be unset here.

nijel (commented):
I've commented on some issues in the code. Can you please add testcase for this backend to make sure that the fix is really working?

asencis:
> I've commented on some issues in the code. Can you please add testcase for this backend to make sure that the fix is really working?

@nijel Ah apologies. I thought a test case existed for every backend, but on a further look - one for orcid never has. Will add this in. I'll also push up the suggested changes.

social_core/backends/orcid.py
nijel: This still applies - in case `orcid_identifier` is None (or empty), `username` is not set and the code will crash here.

asencis: @nijel username is set to an empty string now...see line 33 of orcid.py:`first_name = last_name = email = username = ''` (correct me if I am wrong, but if username is empty like this - the backend will create a unique one "automagically"?)

nijel: Ah sorry, I've missed that line.

asencis: @nijel No problem. Tests are still ongoing...do we use pytest for this library?

nijel: Yes, it uses pytest, see `tox.ini`.

nijel:
There seems to be something wrong in this PR - the tests which normally run in a minute are not finished in two hours. I'm restarting the builds now to see if that was some temporary issue...

asencis:
> PS: It seems to hang in OrbiOAuth2Test ... maybe it's using some network service which blocks?

@nijel I've only added tests for ORCID backend, not Orbi (close, but not related to our changes).

nijel:
Sorry for not being clear - Orbi is the last one to pass, then the test hangs, so my assumption is that its ORCID that hangs...

social_core/tests/backends/test_orcid.py
```diff
@@ -0,0 +1,25 @@
+import json
+
+from .oauth import OAuth2Test
+
+from social_core.backends.orcid import ORCIDOAuth2
+
+class ORCIDOAuth2Test(OAuth2Test):
+    backend_path = 'social_core.backends.orcid.ORCIDOAuth2'
+    user_data_url = 'https://login.orbi.kr/oauth/user/get'
```
nijel, line 9: This should probably match USER_ID_URL...

asencis: @nijel Ahhhh random. Will take a look.

asencis: > Sorry for not being clear - Orbi is the last one to pass, then the test hangs, so my assumption is that its ORCID that hangs...

Yeh that's probably a fair shout, it's the one alphabetically before ORCID.

asencis:
@nijel I'll run the tests localy now and debug as and when we hit issues.

asencis:
@nijel Can you advise on the tox.ini installation commands for installing all of the required interpreters please...

nijel:
You don't need to use tox and test all Python versions.

Either run tox py38 just for the version you use.

Or install needed deps directly via `pip install -r social_core/tests/requirements.txt` and then run `pytest` with your test file.

social_core/tests/backends/test_orcid.py
asencis: > You don't need to use tox and test all Python versions.
> 
> Either run tox py38 just for the version you use.
> 
> Or install needed deps directly via `pip install -r social_core/tests/requirements.txt` and then run `pytest` with your test file.



> You don't need to use tox and test all Python versions.
> 
> Either run tox py38 just for the version you use.
> 
> Or install needed deps directly via `pip install -r social_core/tests/requirements.txt` and then run `pytest` with your test file.

For both of these actions I'm getting a python wheels error: `ERROR: Could not build wheels for xmlsec which use PEP 517 and cannot be installed directly`

nijel: You will have to install xmlsec headers for that, on Ubuntu/Debian it's called `libxmlsec1-dev`.

asencis:
@nijel I think the tests for ORCID weren't in place before because I don't see a way to accurately test it...you can't create a default "foobar" account with ORCID... please advise on how we can test this?

social_core/tests/backends/test_orcid.py
nijel: This is where you mock the responses of the service, it should not talk to the real service for the tests. For example I've added OAuth2 tests for Dropbox in https://github.com/python-social-auth/social-core/pull/557/files

asencis: @nijel We have looked at the testing functionality from around the app - unfortunately ORCID does not structure things in the same way as other standard oauth backends, to get the ORCID id to query their API we need to make an oauth/userinfo call first. The python-social-auth application is returning a generated username -> so we can not test our particular method unfortunately. Please advise.

asencis: @nijel So we find the following:

```
FAILED social_core/tests/backends/test_orcid.py::ORCIDOAuth2Test::test_login - AssertionError: 'e07051f297fc4282a36bf05cb262b4d8' != '0000-0002-2601-8132'
FAILED social_core/tests/backends/test_orcid.py::ORCIDOAuth2Test::test_partial_pipeline - AssertionError: '6bcf2f9f1e4247e9b6e891d1f08073b0' != '0000-0002-2601-8132'
```

nijel: I think all OAuth backends work more or less same - you do OAauth, get some kind of ID and OAuth token and do additional request to get more information about the user. This can be easily covered by the tests....

nijel, line 9: Start with this, it will mock that URL. 

```suggestion
    user_data_url = ORCIDOAuth2.USER_ID_URL
```

You will probably have to modify the other mock params below (I have not checked that).

asencis: @nijel We're struggling to understand the testing then ... we need more advise on how to test the changes...

asencis: @nijel We have modified these. Will push up amendments.

social_core/tests/backends/test_orcid.py
```diff
@@ -0,0 +1,26 @@
...
+
+from social_core.backends.orcid import ORCIDOAuth2
+
+class ORCIDOAuth2Test(OAuth2Test):
+    backend_path = 'social_core.backends.orcid.ORCIDOAuth2'
+    user_data_url = 'https://orcid.org/oauth/userinfo'
+    expected_username = '0000-0002-2601-8132'
+    access_token_body = json.dumps({
+        'access_token': 'foobar',
+        'token_type': 'bearer'
```
nijel, line 13:
```suggestion
        'token_type': 'bearer',
        'orcid-identifier': {'path': '0000-0002-2601-8132'},
```

This is how you extract username in the backend, so you should inject it here as well (or at least I hope I understand your code). Anyway put here the expected OAuth response which is parsed `by get_user_details`.

nijel:

Issue #570, opened 2021-03-05
BUG: Slack user identity API fails


matdrapeau:
Current user_data is providing access_token as query string while it should be provided as authentication bearer in headers.
As described here:
https://api.slack.com/methods/users.identity

Currently, authentication fails with `invalid_auth` in response.

nijel:
Fixed by https://github.com/python-social-auth/social-core/pull/572

PR #572, opened 2021-03-05
BUGFIX: Slack backend user_data should provide access_token

matdrapeau:
Slack user identity API requires access token to be provided in Bearer headers.
https://api.slack.com/methods/users.identity#arg_token

bug as described here: https://github.com/python-social-auth/social-core/issues/570


nijel:

PR #576, opened 2021-03-16
Add refresh_token to EXTRA_DATA

shawncrawley:
The ArcGIS Rest API for OAuth supports using the refresh_token, so it should be included in EXTRA_DATA by default. See https://developers.arcgis.com/rest/users-groups-and-items/token.htm


nijel:

PR #577, opened 2021-03-18
Set a JWT signature algorithm for the Globus backend to RS512

lukaszlacinski:
When `JWT_ALGORITHMS = [RS256]` was hard coded in [backends/open_id_connect.py](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/open_id_connect.py), web apps with the Globus backend started throwing `social_core.exceptions.AuthTokenError: Token error: Invalid signature`. This PR overwrites JWT_ALGORITHMS in the Globus backend.

### DIFF
--- a/social_core/backends/globus.py
+++ b/social_core/backends/globus.py
@@ -10,6 +10,7 @@ from social_core.backends.open_id_connect import OpenIdConnectAuth
 class GlobusOpenIdConnect(OpenIdConnectAuth):
     name = 'globus'
     OIDC_ENDPOINT = 'https://auth.globus.org'
+    JWT_ALGORITHMS = ['RS256', 'RS512']
     EXTRA_DATA = [
         ('expires_in', 'expires_in', True),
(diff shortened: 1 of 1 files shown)


nijel:
Do you see a good reason for not enabling RS512 on the parent class?

lukaszlacinski:
RS512 can be added to the parent class but I am not sure if it is a good idea to modify the parent class every time when a backend needs to use an algorithm that is not listed in the parent class. RS256 was hardcoded in the parent class to force backend classes to verify the JWT signature.

nijel:
I have no clue how this works, but I expected that adding more secure algorithm would not reduce security.

NickolausDS:
This appears to be fixed in the latest version of social core:
```
social-auth-app-django==5.0.0
social-auth-core==4.1.0
```
Did something change in `4.1.0` to check available algorithms in `id_token_signing_alg_values_supported`?

nijel:

Issue #583, opened 2021-04-10
access_token is not being saved for VKAppOAuth2 backend


desecho:
### Expected behaviour

`access_token` should be saved in extra data.

### Actual behaviour

`access_token` is always empty.

### What are the steps to reproduce this issue?

1. Perform login with `VKAppOAuth2` backend
2. Check the `extra data` field for the newly created user here `/admin/social_django/usersocialauth/`

desecho:
The access token is not actually working. It gives this error - `User authorization failed: access_token was given to another ip address`. I am not sure if it should work at all so I closed the issue.

PR #584, opened 2021-04-10
Fix bug where access_token is not being saved for VKAppOAuth2 backend

desecho:
## Proposed changes

https://github.com/python-social-auth/social-core/issues/583

## Types of changes

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Fixed Slack user identity API call with Bearer headers
+- Fixed bug where `access_token` is not being saved for VKAppOAuth2 backend
 
 
 ## [4.1.0](https://github.com/python-social-auth/social-core/releases/tag/4.1.0) - 2021-03-01
@@ -73,7 +74,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Directly use `access_token` in Azure Tenant backend
-- Support Apple JWT audience 
+- Support Apple JWT audience
 - Update partial session cleanup to remove old token from session too
 - Fetch user email in Okta integration
 - Improve Python 3.9 compatibility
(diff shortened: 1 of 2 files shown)


social_core/backends/oauth.py
```diff
@@ -40,8 +40,9 @@ def extra_data(self, user, uid, response, details=None, *args, **kwargs):
         """Return access_token and extra defined names to store in
         extra_data field"""
         data = super().extra_data(user, uid, response, details, *args, **kwargs)
-        data['access_token'] = response.get('access_token', '') or \
-            kwargs.get('access_token')
+        data['access_token'] = (response.get('access_token', '') or
+            kwargs.get('access_token') or
+            kwargs['request'].get('access_token') if 'request' in kwargs else None)
```
nijel, line 45: This looks wrong. The request is already passed as named arg and should not appearing the kwargs

desecho: From the debugging - args is actually an empty tuple. And kwargs is like this:
```
{'auth': <social_core.backends.vk.VKAppOAuth2 object at 0x7fd18e08dc40>,
 'is_new': False,
 'new_association': False,
 'pipeline_index': 8,
 'request': {'access_token': '...',
             ...},
 'social': <UserSocialAuth: John Doe>,
 'storage': <class 'social_django.models.DjangoStorage'>,
 'strategy': <social_django.strategy.DjangoStrategy object at 0x7fd18e0308b0>,
 'username': 'john'}
```

nijel: Sorry, I misread the method signature...

Anyway looking again at this (and hopefully more carefully) - `OAuthAuth.extra_data` is definitely not the place to address this behavior. The `request` is here present only because `VKAppOAuth2.auth_complete` behaves differently than in the base OAuth2 class.

The proper fix would be to make `VKAppOAuth2.auth_complete` consistent with other OAuth2 backends.

desecho: Sorry, yes, that makes sense. I made a change in `VKAppOAuth2.auth_complete`.

desecho:
The access token is not actually working. It gives this error - `User authorization failed: access_token was given to another ip address`. I am not sure how to make it work or if it should work at all so I will close the MR.

PR #600, opened 2021-06-21
shopify: Update API version

nijel:
The old one is no longer supported, see https://github.com/python-social-auth/social-docs/pull/87

## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers
why we should accept this pull request. If it fixes a bug or resolves a feature
request, be sure to link to that issue.

## Types of changes

## Checklist

your code._

## Other information

### DIFF
--- a/social_core/backends/shopify.py
+++ b/social_core/backends/shopify.py
@@ -22,7 +22,7 @@ class ShopifyOAuth2(BaseOAuth2):
 
     @property
     def shopify_api_version(self):
-        return self.setting('API_VERSION', '2019-04')
+        return self.setting('API_VERSION', '2020-10')
 
     @property
     def shopify_api(self):


social_core/backends/shopify.py
```diff
@@ -22,7 +22,7 @@ class ShopifyOAuth2(BaseOAuth2):
 
     @property
     def shopify_api_version(self):
-        return self.setting('API_VERSION', '2019-04')
+        return self.setting('API_VERSION', '2020-10')
```
agconti, line 25: Hey @nijel, thanks for making this PR! 💖 

This issue will keep cropping up once a year since Shopify [depreciates](https://www.shopify.com/partners/blog/api-deprecation) its quarterly API versions yearly. 

We could take an approach where we bump this version yearly or even make this default to dynamically generated string for the current year and quarter. However, this will break existing app's that upgrade social-auth if they do not also upgrade the python Shopify SDK in step. I think it's worth making users set this value in their own config so that upgrading the Shopify SDK is more obvious and intentional. 

What do you think?

nijel: As there is no 2021 API version yet, I don't think dynamic generation is a way to go. So bumping it manually when needs is probably the best approach.

agconti: I agree, dynamically is definitely not the way to go. Why not require the end-user to set the value?

nijel: The other backends have it defined as well (though it's probably not that much rolling target as here...). For example:

https://github.com/python-social-auth/social-core/blob/0bc3fef3b7a14b6e230bd4b25c1617ee8acaa2e6/social_core/backends/facebook.py#L17

agconti: Ahh okay, got ya. 👍

PR #604, opened 2021-08-07
fix microsoft login error

naritotakizawa:
#605


nijel:

Issue #607, opened 2021-08-21
Twitch OAuth2 backend uses deprecated API


ghost:
### Expected behaviour

The Twitch backend should just work, however, when configured it will error out, as it makes an invalid call to an API that was deprecated and is set to be removed by the end of the year. Any authentication attempts fail with a call to the user information endpoint.

### Actual behaviour

The user should be able to authenticate and the call to the user info endpoint should succeed.

### What are the steps to reproduce this issue?

Input clear steps to reproduce the issue for a maintainer.

1. Use the TwitchOAuth2 backend.
2. Sign in or connect an account to it.
3. Observe the error.

### Any logs, error output, etc?

It will throw an exception.

### Any other comments?

I've a working OpenIdConnect backend in my source tree, I am in the process of writing tests and documentation for it and will be making a pull request for it later today.

bloodwithmilk25:
I made it work for my project, now I will created a pull request for it.
https://github.com/python-social-auth/social-core/pull/619

PR #608, opened 2021-08-21
OpenID Connect backend for Twitch authentication (#607)

spell:
## Proposed changes

This pull request adds a new authentication backend for Twitch using OpenID Connect

## Types of changes

## Checklist

## Other information

There is no documentation included, as I could not find it in the sources. The old backend _should_ be deprecated since the API it uses is also deprecated, but I did not feel comfortable removing it.

### DIFF
--- a/social_core/backends/twitch.py
+++ b/social_core/backends/twitch.py
@@ -3,6 +3,32 @@ Twitch OAuth2 backend, docs at:
     https://python-social-auth.readthedocs.io/en/latest/backends/twitch.html
 """
 from .oauth import BaseOAuth2
+from .open_id_connect import OpenIdConnectAuth
+
+
+class TwitchOpenIdConnect(OpenIdConnectAuth):
+    """Twitch OpenID Connect authentication backend"""
+    name = 'twitch'
+    USERNAME_KEY = 'preferred_username'
+    OIDC_ENDPOINT = 'https://id.twitch.tv/oauth2'
+    DEFAULT_SCOPE = ['openid', 'user:read:email']
+    TWITCH_CLAIMS = '{"id_token":{"email": null,"email_verified":null,"preferred_username":null}}'
+
+    def auth_params(self, state=None):
+        params = super().auth_params(state)
+        # Twitch uses a non-compliant OpenID implementation where the claims must be passed as a param
+        params['claims'] = self.TWITCH_CLAIMS
+        return params
+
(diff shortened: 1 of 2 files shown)


social_core/backends/twitch.py
```diff
@@ -3,6 +3,32 @@
...
+
+    def get_user_details(self, response):
+        return {
+            'username': self.id_token['preferred_username'],
+            'email': self.id_token['email'],
+            # Twitch does not provide this information
+            'fullname': '',
+            'first_name': '',
+            'last_name': '',
+        }
```
nijel, line 31: Can you please add tests for get_user_details? See other tests how it is usually done.

social_core/tests/backends/test_twitch.py
```diff
@@ -1,5 +1,52 @@
...
+from .open_id_connect import OpenIdConnectTestMixin
+
+
+class TwitchOpenIdConnectTest(OpenIdConnectTestMixin, OAuth2Test):
+    backend_path = 'social_core.backends.twitch.TwitchOpenIdConnect'
+    user_data_url = 'https://id.twitch.tv/oauth2/userinfo'
+    issuer = 'https://id.twitch.tv/oauth2'
+    expected_username = 'test_user1'
+    openid_config_body = json.dumps({
+        "authorization_endpoint": "https://id.twitch.tv/oauth2/authorize",
```
nijel, line 12: Please use single quotes to follow existing code style.

spell: Ah, my mistake on that one. I normally use single quotes. I will fix that.

social_core/backends/twitch.py
spell: Very well.

spell:
I've fied the quotes, however, I do not understand how to add a test for `get_user_details()` and will need some assistance or concrete examples I can look at. The tests for this project are rather confusing to me.

nijel:
Adding user_data / user_data_body attributes to the Test class should do that.

spell:
I've looked at that nijel, but I don't really get what that tests. I can literally enter anything I want into the `user_data_body` and it still passes. Someone is going to have to explainto me what exactly it is meant to test.

nijel:
It probably tests that parsing in get_user_data doesn't crash. I haven't looked at that, but I've noticed that that method is not covered by tests...

nijel:

PR #617, opened 2021-09-21
vk: improve code quality

blacktrub:
## Proposed changes

I wanted to improve code quality of a VK back-end, because I had a problem when I wanted to extend it. So I decided it's a good idea to move `vk_api` function to class implementation, so we can use an inheritance with full power.

## Types of changes

## Checklist

your code._

## Other information


social_core/backends/vk.py
```diff
@@ -12,6 +12,46 @@
...
+    API_BASE_URL = 'https://api.vk.com'
+
+    def get_api_url(self, method, data):
+        if 'access_token' in data:
+            return self.API_BASE_URL + '/method/' + method
+        return self.API_BASE_URL + '/api.php'
+            
+    def make_sure_access_token_exists(self, method, data):
+        if 'access_token' in data:
+            return data
```
nijel, line 28: Seems like access_token is always present, maybe the code can be simplified?

blacktrub: I'm afraid it's not, when we call the method from VKAppOAuth2, we don't have access_token.

nijel: Ah, I missed the `isAppUser` call when reviewing, and lack of test coverage there made me think it is not used. Can you add tests for that?

blacktrub: Sure, I'll add.

ypankovych (approved):

social_core/backends/vk.py
nijel: @blacktrub Any chance to get to this? :smile:

PR #620, opened 2021-10-12
Disable stale bot

andersk:
Stale bot is worse than useless for this repository. The idea than an issue automatically becomes “stale” just because another 60 days have passed with *no releases* is absurd. Closing valid reports without fixing them is frustrating to reporters, and I would imagine that repeatedly getting “marked as stale”, “still an issue”, “marked as stale”, “still an issue”, … comment spam is frustrating to maintainers too. The fact that this project is currently short on maintainer time is not relevant—that just makes it even more important to leave issues open so they can be dealt with in the future when there is time. Nobody is served by trying to pretend there are no bugs.


nijel:
Indeed, it is probably more annoying than useful without doing some review of the issues.

PR #633, opened 2021-11-17
Replace if statement with if expression

yezz123:
## Proposed changes

Python's conditional expression syntax is its version of the ternary operator. Using this syntax is definitely more concise, but it is one of the more controversial refactorings (along with list comprehensions). Some coders dislike these expressions and find them slightly harder to parse than writing them out fully.

## Types of changes

- [x] Refactoring (no functional changes, no API changes)

## Checklist

### DIFF
--- a/social_core/utils.py
+++ b/social_core/utils.py
@@ -108,28 +108,25 @@ def sanitize_redirect(hosts, redirect_to):
 
 def user_is_authenticated(user):
     if user and hasattr(user, 'is_authenticated'):
-        if callable(user.is_authenticated):
-            authenticated = user.is_authenticated()
-        else:
-            authenticated = user.is_authenticated
+        return (
+            user.is_authenticated()
+            if callable(user.is_authenticated)
+            else user.is_authenticated
+        )
+
     elif user:
-        authenticated = True
+        return True
     else:
-        authenticated = False
-    return authenticated
+        return False
 
 
 def user_is_active(user):
     if user and hasattr(user, 'is_active'):
-        if callable(user.is_active):
-            is_active = user.is_active()
-        else:
-            is_active = user.is_active
+        return user.is_active() if callable(user.is_active) else user.is_active
     elif user:
-        is_active = True
+        return True
     else:
-        is_active = False
-    return is_active
+        return False
 
 
 # This slugify version was borrowed from django revision a61dbd6
@@ -235,10 +232,9 @@ def setting_url(backend, *names):
     for name in names:
         if is_url(name):
             return name
-        else:
-            value = backend.setting(name)
-            if is_url(value):
-                return value
+        value = backend.setting(name)
+        if is_url(value):
+            return value
 
 
 def handle_http_errors(func):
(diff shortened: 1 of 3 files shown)


social_core/utils.py
```diff
@@ -108,28 +108,25 @@ def sanitize_redirect(hosts, redirect_to):
...
+        return (
+            user.is_authenticated()
+            if callable(user.is_authenticated)
+            else user.is_authenticated
+        )
+
     elif user:
-        authenticated = True
+        return True
     else:
```
nijel, line 119: Else not needed here.

social_core/utils.py
```diff
@@ -108,28 +108,25 @@ def sanitize_redirect(hosts, redirect_to):
...
     if user and hasattr(user, 'is_active'):
-        if callable(user.is_active):
-            is_active = user.is_active()
-        else:
-            is_active = user.is_active
+        return user.is_active() if callable(user.is_active) else user.is_active
     elif user:
-        is_active = True
+        return True
     else:
```
nijel, line 128: Else not needed here.

social_core/utils.py
```diff
@@ -235,10 +232,9 @@ def setting_url(backend, *names):
     for name in names:
         if is_url(name):
             return name
-        else:
-            value = backend.setting(name)
-            if is_url(value):
-                return value
+        value = backend.setting(name)
+        if is_url(value):
+            return value
```
nijel, line 237: I'd add explicit return here:

```suggestion
            return value
        return None
```

nijel (commented):
Thanks! Looks good, but I see some more code improvements.

social_core/utils.py
yezz123: I think its better if we let it, in the test case we gonna back to it

nijel: I mean that the "else" statement is not needed, not the code.

GitHub doesn't allow me to make the proper suggestion over removed lines, but basically this (and removing the next `return False` statement):

```suggestion
    else:
    return False
```

yezz123: Fixed @nijel you could run the tests

social_core/utils.py
```diff
@@ -235,10 +230,10 @@ def setting_url(backend, *names):
...
+        return None
```
nijel, line 236: Shouldn't this be outside the loop?

denisSurkov:
@yezz123 , do you plan to continue work on the pr?

yezz123:
> @yezz123 , do you plan to continue work on the pr?

Yes, I was busy just with some work I just merge this https://github.com/python-social-auth/social-core/pull/633/commits/f18da50c093ab31d2b99ef5552b84b90147c032e if you have any improvement in the code that's would be great

gersona:
@yezz123 is there more work to do on this PR beside the conflicts to be resolved ?

nijel:
@gersona I think another review that it doesn't change the behavior would be a good start.

social_core/utils.py
```diff
@@ -116,22 +116,15 @@ def user_is_authenticated(user):
         if callable(user.is_authenticated):
             authenticated = user.is_authenticated()
         else:
-            authenticated = user.is_authenticated
-    elif user:
-        authenticated = True
-    else:
-        authenticated = False
-    return authenticated
+            return True
```
nijel, line 119: This no longer checks whether user is authenticated.

PR #634, opened 2021-11-29
print error in apple backend

cuococarlos:
## Proposed changes
I propose add new log in apple backend, because AuthFailed raise in [this line](https://github.com/python-social-auth/social-core/blob/eb2010b542e7a5963caca291cce8a4caee23a09b/social_core/backends/apple.py#L132) , I don't know  which is error generate in this point

## Types of changes

## Checklist

your code._

Screenshots:
image

### DIFF
--- a/social_core/backends/apple.py
+++ b/social_core/backends/apple.py
@@ -128,8 +128,8 @@ class AppleIdAuth(BaseOAuth2):
                 audience=self.get_audience(),
                 algorithms=['RS256'],
             )
-        except PyJWTError:
-            raise AuthFailed(self, 'Token validation failed')
+        except PyJWTError as error:
+            raise AuthFailed(self, 'Token validation failed by: ' + error.__str__())
 
         return decoded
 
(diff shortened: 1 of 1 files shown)


social_core/backends/apple.py
```diff
@@ -128,8 +128,8 @@ def decode_id_token(self, id_token):
                 audience=self.get_audience(),
                 algorithms=['RS256'],
             )
-        except PyJWTError:
-            raise AuthFailed(self, 'Token validation failed')
+        except PyJWTError as error:
+            raise AuthFailed(self, 'Token validation failed by: ' + error.__str__())
```
nijel, line 132: Using f-string will make the code more readable:

```suggestion
            raise AuthFailed(self, f'Token validation failed by {error}')
```

Still, I'm not convinced whether it's desired to expose this to the user. Won't PyJWT leak credentials in the exceptions?

cuococarlos: @nijel thanks for revision. In my test the credentials don't leak.  Only the cause of the error is printed. I use social-core integred with [drf-social-oauth](https://github.com/wagnerdelima/drf-social-oauth2) in case you're wondering why the REST call in the log

image

PD: Sorry for my basic english

nijel:

PR #613, opened 2021-09-13
Update vk backend to the 5.131

ffedoroff:
Update vk backend to the latest api. More info here: https://vk.com/dev/versions

## Proposed changes

Just bump to new version.

## Types of changes

Please check the type of change your PR introduces:

- [ ] Release (new release request)
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (PEP8, lint, formatting, renaming, etc)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build related changes (build process, tests runner, etc)
- [ ] Other (please describe):

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask. We're here to
help! This is simply a reminder of what we are going to look for before merging
your code._

- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works

## Other information

Just bump to new version.

PR #648, opened 2021-12-28
Version up of VK on invoking auth method

PixxxeL:
## Proposed changes

Up version to 5.81 because AuthException at `.../complete/vk-oauth2/` is raised: "Invalid request: versions below 5.81 are deprecated. Version param should be passed as "v". "version" param is invalid and not supported. For more information go to https://vk.com/dev/constant_version_updates"

## Types of changes

* Bugfix (non-breaking change which fixes an issue)

### DIFF
--- a/social_core/backends/vk.py
+++ b/social_core/backends/vk.py
@@ -178,7 +178,7 @@ def vk_api(backend, method, data):
         http://goo.gl/yLcaa
     """
     # We need to perform server-side call if no access_token
-    data['v'] = backend.setting('API_VERSION', '5.131')
+    data['v'] = backend.setting('API_VERSION', '5.81')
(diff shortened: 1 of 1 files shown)


social_core/backends/vk.py
```diff
@@ -178,7 +178,7 @@ def vk_api(backend, method, data):
         http://goo.gl/yLcaa
     """
     # We need to perform server-side call if no access_token
-    data['v'] = backend.setting('API_VERSION', '5.131')
+    data['v'] = backend.setting('API_VERSION', '5.81')
```
nijel, line 181: You are changing this to *older* version. See also https://github.com/python-social-auth/social-core/pull/613 or https://github.com/python-social-auth/social-core/pull/615

PR #661, opened 2022-02-02
Assume RSA256 if key "alg" not set in open_id_connect.py

spelufo:
This change, proposed [here](https://github.com/mpdavis/python-jose/pull/171#issuecomment-574311477), fixed Signature verification errors we had with ping identity.

### DIFF
--- a/social_core/backends/open_id_connect.py
+++ b/social_core/backends/open_id_connect.py
@@ -160,7 +160,7 @@ class OpenIdConnectAuth(BaseOAuth2):
         for key in keys:
             if kid is None or kid == key.get('kid'):
                 if 'alg' not in key:
-                    key['alg'] = 'RSA256'
+                    key['alg'] = self.JWT_ALGORITHMS[0]
                 rsakey = jwk.construct(key)
                 message, encoded_sig = id_token.rsplit('.', 1)
                 decoded_sig = base64url_decode(encoded_sig.encode('utf-8'))
--- a/social_core/backends/open_id_connect.py
+++ b/social_core/backends/open_id_connect.py
@@ -159,6 +159,8 @@ class OpenIdConnectAuth(BaseOAuth2):
 
         for key in keys:
             if kid is None or kid == key.get('kid'):
+                if 'alg' not in key:
+                    key['alg'] = 'RSA256'
                 rsakey = jwk.construct(key)
                 message, encoded_sig = id_token.rsplit('.', 1)
                 decoded_sig = base64url_decode(encoded_sig.encode('utf-8'))


social_core/backends/open_id_connect.py
```diff
@@ -159,6 +159,8 @@ def find_valid_key(self, id_token):
 
         for key in keys:
             if kid is None or kid == key.get('kid'):
+                if 'alg' not in key:
+                    key['alg'] = 'RSA256'
```
dylantack, line 163: My identity provider has this issue also. Instead of a fixed default, how about using the first of JWT_ALGORITHMS?

I agree that we don't want to consider the JWT's alg flag since it's so easily [misused](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/). 

```suggestion
                    key['alg'] = self.JWT_ALGORITHMS[0]
```

dylantack, line 163: or more strictly:

```suggestion
                if 'alg' not in key and len(JWT_ALGORITHMS) == 1:
                    key['alg'] = self.JWT_ALGORITHMS[0]
```

davidhalter:
As far as I can tell, there is no default for the `alg` header ([link](https://www.rfc-editor.org/rfc/rfc7515.html#section-4.1.1))

> 4.1.1.  "alg" (Algorithm) Header Parameter

>   [...] This Header Parameter MUST be present and MUST be understood and processed by implementations.

So I would probably close this, because this is not according to spec.

davidhalter:
OK, I'm realizing that I'm also not sure about the specs anymore, because there's a bit of confusion around JWK/JOSE. Please quote the correct spec and I'll reconsider.

dylantack:
@davidhalter 

From [RFC 7517, section-4.4:](https://datatracker.ietf.org/doc/html/rfc7517#section-4.4):

> The "alg" (algorithm) parameter identifies the algorithm intended for
>    use with the key.  The values used should either be registered in the
>    IANA "JSON Web Signature and Encryption Algorithms" registry
>    established by [[JWA](https://datatracker.ietf.org/doc/html/rfc7517#ref-JWA)] or be a value that contains a Collision-
>    Resistant Name.  The "alg" value is a case-sensitive ASCII string.
>    **Use of this member is OPTIONAL.**

"alg" is required in the JWT, but optional in the JWKS.

At first I thought we could use the JWT's "alg" field in this case (`jwt.get_unverified_header(id_token).get('alg')` ), then realized this is unsafe -- you [can't trust](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/) the JWT's alg property.

I think what I suggested above (use the JWT_ALGORITHMS[0] as a default if, and only if, there's a single allowed alg) is safe, and within the spec.

davidhalter:
Thanks for your research. I'm still a bit unsure what the correct solution is here, because the RFC does not tell me much. However defaulting to either `RS256` or using the first value of `JWT_ALGORITHMS` should be fine I guess.

What do you think @spelufo?

spelufo:
@cultcom , in the issue I linked before, refered to this part of the spec:

https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation

> 7. The alg value SHOULD be the default of RS256 or the algorithm sent by the Client in the id_token_signed_response_alg parameter during Registration.

dylantack:
@spelufo That's interesting, I hadn't noticed that part of the OIDC spec. I think we still need at least the option for an implementor to override the default. Based on the spec you quoted, other algs are possible -- RS256 is a "SHOULD" (not a "MUST"), and clients could register with a different alg in id_token_signed_response_alg (for example ES384).

JWT_ALGORITHMS is _already_ populated with a default of `['RS256']` so I don't think this exposes more complexity to the implementor. Also if a future version of social-auth supported discovery of JWT_ALGORITHMS (`id_token_signing_alg_values_supported` is a required property in `/.well-known/openid-configuration`), it would automatically pick up the correct default.

davidhalter:
I also saw that part of the doc, but I wonder if this even applies. Couldn't this be what the JOSE header should specify? All in all I'm really confused what the correct option should be here. I tried to lookup other implementations and see what they do, but could not find that part of the implementation...

So I will probably just merge.

davidhalter:
Thanks guys!

PR #666, opened 2022-02-15
Fix setting fullname for ORCID

hmpf:
Originally, fullname was set to `response['person']['name']`, which in
ORCID is a large json object. Everywhere else, fullname is assumed to be
a string. It cannot be used directly as a fullname but needs converting.

## Proposed changes

The patch ensures that fullname is a string, in a similar way to how other backends do it.

## Types of changes

## Checklist

(I had to install libxmlsec1-dev manually on debian in order to get tox to run.)

### DIFF
--- a/social_core/backends/orcid.py
+++ b/social_core/backends/orcid.py
@@ -70,11 +70,11 @@ class ORCIDOAuth2(BaseOAuth2):
         if person:
             name = person.get('name')
 
-            fullname = name
-
             if name:
                 first_name = name.get('given-names', {}).get('value', '')
                 last_name = name.get('family-name', {}).get('value', '')
+                fullname = first_name + ' ' + last_name
+                fullname = fullname.strip()
 
             emails = person.get('emails')
             if emails:


social_core/backends/orcid.py
```diff
@@ -70,11 +70,11 @@ def get_user_details(self, response):
         if person:
             name = person.get('name')
 
-            fullname = name
-
             if name:
                 first_name = name.get('given-names', {}).get('value', '')
                 last_name = name.get('family-name', {}).get('value', '')
+                fullname = first_name + ' ' + last_name
+                fullname = fullname.strip()
```
nijel, line 77: Can you please look why this is not covered by tests? The user data payload seems to be present in the test case, but it doesn't seem to be used.

social_core/backends/orcid.py
```diff
@@ -70,12 +70,13 @@ def get_user_details(self, response):
         if person:
             name = person.get("name")
 
-            fullname = name
-
             if name:
                 first_name = name.get("given-names", {}).get("value", "")
+                last_name = name.get("family-name", {}).get("value", "")
```
nijel, line 75:
```suggestion
```

nijel:

PR #668, opened 2022-02-18
Add backend for Hashicorp Vault

candlerb:
## Proposed changes

This is a new backend for the [OIDC Provider](https://www.vaultproject.io/docs/secrets/identity/oidc-provider) in Hashicorp Vault 1.9+

## Types of changes

Please check the type of change your PR introduces:

- [ ] Release (new release request)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Code style update (PEP8, lint, formatting, renaming, etc)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build related changes (build process, tests runner, etc)
- [ ] Other (please describe):

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask. We're here to
help! This is simply a reminder of what we are going to look for before merging
your code._

- [X] Lint and unit tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works

(Not sure what linting you want, but flake8 passes)

## Other information

Tested with [Netbox](https://github.com/netbox-community/netbox/discussions/8579#discussioncomment-2200174)

NOTE 1: I had to override the base `OpenIdConnectAuth` class to do HTTP Basic Auth when talking to the token endpoint. Maybe it would be better if functionality were moved into the base class itself, since it can make use of the information exposed in the OIDC discovery response:

```
      'token_endpoint_auth_methods_supported': [ 'client_secret_basic' ],
```

NOTE 2: I turned OIDC_ENDPOINT into a property, since this is something the user has to configure to point to their own Vault server.  Maybe it would be better to turn this into a 'setting' on the `OpenIdConnectAuth` base class.

NOTE 3: With those two changes, I would have been able to instantiate the OpenIdConnectAuth class directly - except that it would still need to have a 'name' property.  If you set 'name' to 'oidc' here, then it would be usable as a generic OpenID Connect backend.  (I made one other change to pick up the 'groups' claim, which is non-standard but commonly implemented)

PR #669, opened 2022-02-22
Add backend for Hashicorp Vault (alternative approach)

candlerb:
## Proposed changes

This is a more generic solution to PR #668 - for consideration.

In this PR, the OpenIdConnectAuth class is enhanced so that it can be instantiated directly, as a "generic" OIDC connector.

To enable this, it also makes the OAuth2 authentication mechanism configurable (i.e. form post or HTTP Basic), and uses OIDC discovery to select this automatically.

Vault then becomes an empty subclass of OpenIdConnectAuth, since it's an out-of-the-box OIDC provider.

## Types of changes

I believe this *shouldn't* break existing code, and all tests pass, but it would be wise to check the other OpenIdConnectAuth subclasses just in case.

## Checklist

your code._

## Other information

This PR also includes a fix for okta test which surfaced as part of this change. I believe this should only fix the test, not the actual behavior of the okta backend.

### DIFF
--- a/social_core/backends/open_id_connect.py
+++ b/social_core/backends/open_id_connect.py
@@ -26,7 +26,15 @@ class OpenIdConnectAuth(BaseOAuth2):
     """
     Base class for Open ID Connect backends.
     Currently only the code response type is supported.
+
+    It can also be directly instantiated as a generic OIDC backend.
+    To use it you will need to set at minimum:
+
+    SOCIAL_AUTH_OIDC_OIDC_ENDPOINT = 'https://.....'  # endpoint without /.well-known/openid-configuration
+    SOCIAL_AUTH_OIDC_KEY = '<client_id>'
+    SOCIAL_AUTH_OIDC_SECRET = '<client_secret>'
     """
+    name = 'oidc'
     # Override OIDC_ENDPOINT in your subclass to enable autoconfig of OIDC
     OIDC_ENDPOINT = None
     ID_TOKEN_MAX_AGE = 600
@@ -37,46 +45,58 @@ class OpenIdConnectAuth(BaseOAuth2):
     REVOKE_TOKEN_METHOD = 'GET'
     ID_KEY = 'sub'
     USERNAME_KEY = 'preferred_username'
+    JWT_ALGORITHMS = ['RS256']
+    JWT_DECODE_OPTIONS = dict()
+    # When these options are unspecified, server will choose via openid autoconfiguration
     ID_TOKEN_ISSUER = ''
     ACCESS_TOKEN_URL = ''
     AUTHORIZATION_URL = ''
     REVOKE_TOKEN_URL = ''
     USERINFO_URL = ''
     JWKS_URI = ''
-    JWT_ALGORITHMS = ['RS256']
-    JWT_DECODE_OPTIONS = dict()
+    TOKEN_ENDPOINT_AUTH_METHOD = ''
 
     def __init__(self, *args, **kwargs):
         self.id_token = None
         super().__init__(*args, **kwargs)
 
     def authorization_url(self):
-        return self.AUTHORIZATION_URL or \
+        return self.setting('AUTHORIZATION_URL', self.AUTHORIZATION_URL) or \
             self.oidc_config().get('authorization_endpoint')
 
     def access_token_url(self):
-        return self.ACCESS_TOKEN_URL or \
+        return self.setting('ACCESS_TOKEN_URL', self.ACCESS_TOKEN_URL) or \
             self.oidc_config().get('token_endpoint')
 
     def revoke_token_url(self, token, uid):
-        return self.REVOKE_TOKEN_URL or \
+        return self.setting('REVOKE_TOKEN_URL', self.REVOKE_TOKEN_URL) or \
             self.oidc_config().get('revocation_endpoint')
 
     def id_token_issuer(self):
-        return self.ID_TOKEN_ISSUER or \
+        return self.setting('ID_TOKEN_ISSUER', self.ID_TOKEN_ISSUER) or \
             self.oidc_config().get('issuer')
 
     def userinfo_url(self):
-        return self.USERINFO_URL or \
+        return self.setting('USERINFO_URL', self.USERINFO_URL) or \
             self.oidc_config().get('userinfo_endpoint')
 
     def jwks_uri(self):
-        return self.JWKS_URI or \
+        return self.setting('JWKS_URI', self.JWKS_URI) or \
             self.oidc_config().get('jwks_uri')
+
+    def use_basic_auth(self):
+        method = self.setting('TOKEN_ENDPOINT_AUTH_METHOD', self.TOKEN_ENDPOINT_AUTH_METHOD)
+        if method:
+            return method == 'client_secret_basic'
+        methods = self.oidc_config().get('token_endpoint_auth_methods_supported', [])
+        return not methods or 'client_secret_basic' in methods
+
+    def oidc_endpoint(self):
+        return self.setting('OIDC_ENDPOINT', self.OIDC_ENDPOINT)
 
     @cache(ttl=86400)
     def oidc_config(self):
-        return self.get_json(self.OIDC_ENDPOINT +
+        return self.get_json(self.oidc_endpoint() +
                              '/.well-known/openid-configuration')
 
     @cache(ttl=86400)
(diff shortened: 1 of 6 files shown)


davidhalter (changes requested):
Thanks for trying to contribute @candlerb 

I'm personally against this big list of changes that affect everything a lot of providers. I think we should go one step smaller and deal with individual problems as they arise. What is the issue for Vault here? What are you trying to solve?

The project is in maintenance mode and I feel like changes like this one are dangerous if there isn't a clear benefit to them. I won't merge for now and I would ask other maintainers not to merge as without discussing this first.

candlerb:
Now I remember the other problem. In `OpenIdConnectAuth `, `OIDC_ENDPOINT` is a class variable, not a setting.

You can spoof it by turning it into a property (which is what I did in #668 to minimise other changes):

```
    @property
    def OIDC_ENDPOINT(self):
        return self.setting('OIDC_ENDPOINT')
```

This works, but it seems pretty ugly to me. It would probably be better just to have a fresh `GenericOpenIdConnectProvider` which doesn't inherit from `OpenIdConnectAuth` at all, even though that involves duplicating code.

The nearest existing backend I found with configurable endpoint URL is the [Keycloak backend](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/keycloak.py), which gets around this problem because it doesn't subclass from `OpenIdConnectAuth`, it subclasses `BaseOAuth2`. However that means it loses all the auto-discovery: you can't just point at Keycloak's base url and pick up `.well-known/openid-configuration`. Instead you have to configure a load of settings by hand (`PUBLIC_KEY`, `AUTHORIZATION_URL`, `ACCESS_TOKEN_URL`)

davidhalter:
I have personally had the same issue. I was able to work around that by using a function that generates those classes and writes them to `globals()`. It's really ugly, but in general it's still the right approach with the code we currently have. I originally also wanted to go with the approach you are taking, but ultimately it's probably the wrong approach for what we have here.

If you are just writing an OpenID Connect compatible software that only needs OIDC, I would probably recommend to not use this library, because it's not very well maintained. There are probably other libraries that only do OpenID Connect and are therefore more correct when it comes to the standard.

In general I'm just worried about touching the OIDC pieces here, because they are already not that well maintained and almost nobody around here understands OIDC on a good level. I certainly do not, because it's just hard to understand and I'm kind of maintaining that part :-/.

candlerb:
> If you are just writing an OpenID Connect compatible software that only needs OIDC, I would probably recommend to not use this library, because it's not very well maintained. There are probably other libraries that only do OpenID Connect and are therefore more correct when it comes to the standard.

It's the other way round: I wanted to take an application which already has python-social-auth integration ([Netbox](https://github.com/netbox-community/netbox/releases/tag/v3.1.0)), and connect it to an OIDC provider.

Netbox does offer an alternative way of doing this, via [remote authentication](https://netbox.readthedocs.io/en/stable/configuration/remote-authentication/) with HTTP headers, so I can instead use an Apache reverse proxy with mod_auth_openidc. I just thought that in fixing this use case for Netbox, it would be nice to make it available to other consumers of python-social-auth.

nijel:
I don't see anything breaking in this (though my OpenID Connect experience is very limited). I think having generic OpenID Connect provider is good for reasons described by @candlerb - we use python-social-auth at Weblate and users do not have the choice for the authentication library (I agree that there might be better choices out there these days), but they want to have the choice in the authentication service. Therefore, I'd go and merge this.

It would be great, if @candlerb could prepare matching docs pull request in https://github.com/python-social-auth/social-docs/

davidhalter:
Since I feel very much responsible for OpenID Connect stuff, please do not merge (otherwise I might have to revert). I will try to review some things again next week. My feeling still is that this is a bad approach (and probably breaks backwards compatibility), but I do not have the time to explain right now.

nijel:
I've just gone again through this patch and I don't see anything breaking compatibility there. I think approach in having generic OpenID Connect backend is desired. For example [IdentityServer](https://github.com/DuendeSoftware/IdentityServer) could be integrated this way also.

davidhalter:
I don't see how this is generic. If I wanted a generic OpenID Connect backend, I would want multiple ones. This is pretty common actually. The current approach is already generic, you just have to instantiate a class.

I still have to review the code better, though.

nijel:
Yes, it's limited to a single OpenID Connect provider, but this is really what I'm looking for. Our users typically have internal authentication service and want to use that. Being able to specify that only in configuration without implementing a subclass for every service would be great.

davidhalter:
> Yes, it's limited to a single OpenID Connect provider, but this is really what I'm looking for. Our users typically have internal authentication service and want to use that.

I feel like the difference of inheriting from a class and setting a few attributes on it versus setting a few variables in a global namespace is minimal. Both of these ways can be documented, will be around the same line count and with inheritance it is even possible to use multiple OIDC backends if that is ever wanted.

-------------

@candlerb

I kind of understand your approach with managing variables better now. Is there a way you could provide me with a way where in the OIDC standard you were looking up the basic auth stuff? I feel like the big problem is that we currently do not support that and the OIDC standard is really hard to get right...

candlerb:
> Is there a way you could provide me with a way where in the OIDC standard you were looking up the basic auth stuff? I feel like the big problem is that we currently do not support that and the OIDC standard is really hard to get right...

Sure. When you want to exchange a code for an ID token (at the Token endpoint), OIDC provides two ways of the client authenticating itself: via HTTP Basic Authentication, or via client_id and client_secret fields in the POST request body.

This is described in [section 3.1.3.2](https://openid.net/specs/openid-connect-core-1_0.html#TokenRequestValidation) and [section 9](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) of the OIDC Core 1.0 specification.

This in turn refers to [section 2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1) of the OAuth 2.0 specification, RFC 6749. Here it is made explicit that the authorization server *must* implement HTTP Basic Auth, and *may* also implement the POST body fields (but the latter is *not recommended*).

By supporting only basic auth but not POST, Vault is compliant to the standard. However, prior to this patch, PSA's oauth.py (and hence also open_id_connect.py) implements only the POST option.

There are some subclasses of BaseOAuth2 which override some methods to do basic auth - see for example RedditOAuth2.

social_core/backends/open_id_connect.py
```diff
@@ -45,38 +53,49 @@ class OpenIdConnectAuth(BaseOAuth2):
     JWKS_URI = ''
     JWT_ALGORITHMS = ['RS256']
     JWT_DECODE_OPTIONS = dict()
+    TOKEN_ENDPOINT_AUTH_METHOD = ''
```
davidhalter, line 56: Shouldn't this be by default `'client_secret_post'`?

davidhalter (commented):
Sorry for the delay again. I'm pretty busy and this is not easy to understand.

I feel like I understand the OpenID Connect part now, thanks for linking the relevant documents.

I feel like since @nijel wants the settings changes, this is also fine (although I personally prefer the subclassing stuff). Is there documentation needed @nijel ?

social_core/backends/open_id_connect.py
candlerb: Agreed. The logic says "anything other than client_secret_basic is client_secret_post" (hence the boolean function `use_basic_auth` in oauth.py), but it doesn't harm to document it properly.

nijel:
Yes, documentation for new backends is definitely needed in https://github.com/python-social-auth/social-docs/

candlerb:
I made some docs a while ago here: https://github.com/python-social-auth/social-docs/compare/master...candlerb:candlerb/oidc

I didn't make a PR as it was unclear whether these oidc changes were accepted, either in current form or differently.

social_core/backends/open_id_connect.py
davidhalter: Ok, I guess I agree. Can you maybe just copy paste your own comment `unspecified, let the server choose via openid autoconfiguration` and add that as a comment? I guess in that case it's clear why it's empty.

candlerb: I guess so, although the same also applies to all these parameters:

```
    ID_TOKEN_ISSUER = ''
    ACCESS_TOKEN_URL = ''
    AUTHORIZATION_URL = ''
    REVOKE_TOKEN_URL = ''
    USERINFO_URL = ''
    JWKS_URI = ''
```

candlerb: I have added a comment covering all these options

davidhalter: Awesome, thanks!

davidhalter (approved):

davidhalter:
To me this looks good now, thanks a lot. I think @nijel should approve all the documentation stuff, I have no idea about that.

nijel:
I've made the PR for the docs here: https://github.com/python-social-auth/social-docs/pull/109

nijel:

PR #670, opened 2022-03-02
Fix for Keycloak backend: remove redirect_state

candlerb:
## Proposed changes

Set `REDIRECT_STATE = False` in keycloak backend.

Without this fix, PSA appends ?redirect_state=XXXX to the redirect_uri parameter value passed to the auth endpoint, and Keycloak rejects this with "Invalid parameter: redirect_uri"

Tested with:
- keycloak 17.0.0
- social-auth-core==4.2.0 plus this patch

## Types of changes

- [?] Breaking change (fix or feature that would cause existing functionality to not work as expected)

It is possibly a "Breaking change", if this backend ever worked before.  But it definitely doesn't with current Keycloak.

## Checklist

your code._

## Other information

User setting `SOCIAL_AUTH_KEYCLOAK_REDIRECT_STATE = False` doesn't fix this, hence as far as I can tell it's necessary to do a code change.

If there are concerns about backwards compatibility / breaking change, this could be turned into a `setting()` that could be overridden by the user.

### DIFF
--- a/social_core/backends/keycloak.py
+++ b/social_core/backends/keycloak.py
@@ -88,6 +88,7 @@ class KeycloakOAuth2(BaseOAuth2):  # pylint: disable=abstract-method
     name = 'keycloak'
     ID_KEY = 'username'
     ACCESS_TOKEN_METHOD = 'POST'
+    REDIRECT_STATE = False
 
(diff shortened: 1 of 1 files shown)


nijel:

PR #676, opened 2022-03-24
add uffd oauth2 backend

davidc:
## Proposed changes

Adds a new backend to support uffd single sign on - https://git.cccv.de/uffd/uffd

As there are multiple uffd installations, the base URL is configured through config not code.

## Types of changes

## Checklist

your code._

### DIFF
--- a/social_core/backends/uffd.py
+++ b/social_core/backends/uffd.py
@@ -1,0 +1,51 @@
+from urllib.parse import urlencode
+
+from .oauth import BaseOAuth2
+
+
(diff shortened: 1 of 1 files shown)


davidc:
> Can you please add tests?

I'm not really sure how I can do that since uffd instances are private

nijel:
The testing doesn't happen against real service, the responses are mocked. It does mostly test processing of user data, see for example https://github.com/python-social-auth/social-core/blob/master/social_core/tests/backends/test_coinbase.py

PR #677, opened 2022-03-28
Allow per-backend user pipeline settings

dehnert:
According to https://python-social-auth.readthedocs.io/en/latest/configuration/settings.html, "All settings can be defined per-backend by adding the backend name to the setting name, like SOCIAL_AUTH_TWITTER_LOGIN_URL". This changes user.py to actually use per-backend settings for options like USERNAME_IS_FULL_EMAIL.

Note that get_username is always called with a backend, but user_details isn't, so a different version of `setting' is needed for the two.

## Types of changes

## Checklist

your code._

- [X?] Lint and unit tests pass locally with my changes -- running tox reports an error in test_dummy both before and after my change

## Other information

### DIFF
--- a/social_core/pipeline/user.py
+++ b/social_core/pipeline/user.py
@@ -11,17 +11,17 @@ def get_username(strategy, details, backend, user=None, *args, **kwargs):
     storage = strategy.storage
 
     if not user:
-        email_as_username = strategy.setting("USERNAME_IS_FULL_EMAIL", False)
-        uuid_length = strategy.setting("UUID_LENGTH", 16)
+        email_as_username = backend.setting("USERNAME_IS_FULL_EMAIL", False)
+        uuid_length = backend.setting("UUID_LENGTH", 16)
         max_length = storage.user.username_max_length()
-        do_slugify = strategy.setting("SLUGIFY_USERNAMES", False)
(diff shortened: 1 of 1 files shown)


dehnert:
I'm confused that CI appears to have still not run.

nijel:
Can you please prepare a matching pull request for the documentation? https://github.com/python-social-auth/social-docs/

dehnert:
I think this change just makes the code match the already-existing documentation -- see https://github.com/python-social-auth/social-docs/blame/master/docs/configuration/settings.rst#L18, "All settings can be defined per-backend". Can you clarify what you're looking for in terms of doc changes?

dehnert:
I'm still running with this patch on one of my sites -- it looks like this PR has a conflict now, but (glancing at the conflict) it doesn't look hard to resolve, or like the conflict is from this issue having been fixed. I can go back and update this PR, but @nijel , can you elaborate on what you were looking for in terms of added docs, so I can resolve that too?

nijel:
I most likely missed your previous reply, it indeed seems that this PR makes it match the documentation.

dehnert:
I've rebased, if you want to take a look and maybe merge it.

nijel:

PR #683, opened 2022-05-08
By default, set the fullname from first+last and vice-versa

### LINKED ISSUE #678: Use fullname and first_name/last_name ~interchangeably
### Expected behaviour

When the SAML backend supplies a `fullname`, my Django user should have `first_name` and `last_name` set: if `fullname` is set but not `first_name`/`last_name`, PSA should (optionally?) generate a `first_name` and `last_name` automatically (my inclination is something like `first_name, _space, last_name = fullname.rpartition(' ')`). If there are systems (that is, non-Django web frameworks) that want a `fullname`, it should do the opposite transformation when `fullname` is missing.

### Actual behaviour

My Django user has neither first_name nor last_name.

Input clear steps to reproduce the issue for a maintainer.

1. Use social_django
2. Configure the SAML backend to return a fullname, but not a first_name or last_name
3. Log in using SAML
4. Observe that the Django user created doesn't have a first_name/last_name set

N/A

### Any other comments?

Unfortunate I don't understand how data flows through PSA well enough to supply a patch.

I suspect that this ideally wants to be configurable (on/off and perhaps a custom `fullname`->`first_name`/`last_name` function) per-backend.

dehnert:
Fixes #678.

## Proposed changes

As described in #678 , when the SAML backend supplies a `fullname`, it's desirable for my Django user to have `first_name` and `last_name` set. This changes PSA so if `fullname` is set (in, eg, a SAML response) but not `first_name`/`last_name`, by default it generates a `first_name` and `last_name` automatically. Similarly, it does the reverse transformation when `fullname` is missing. IMO this is a net improvement, but obviously there's questions like "should this be enabled by default", "should it be in the default pipeline", etc. -- I'm happy to tweak this to make it fit the team's design goals better. Assuming the concept makes sense, I can write docs.

## Types of changes

## Checklist

your code._

- [X?] Lint and unit tests pass locally with my changes -- running tox reports an error in test_dummy both before and after my change

## Other information


social_core/pipeline/social_auth.py
```diff
@@ -5,6 +5,21 @@ def social_details(backend, details, response, *args, **kwargs):
...
 
 
+def social_names(backend, details, response, *args, **kwargs):
+    # If first+last are both missing, populate from full
+    if details.get('fullname') and backend.setting('FIRSTLAST_FROM_FULL', True):
+        if not (details.get('first_name') or details.get('last_name')):
+            first, _space, last = details['fullname'].rpartition(' ')
+            details['first_name'] = first.strip()
+            details['last_name'] = last.strip()
+            print(f"end social_names {details=}")
```
dehnert, line 15: (I'll need to remove this, but I'd like to confirm that the general direction seems good before I do the cleanup.)

dehnert:
Any thoughts on this general feature?

dehnert:
Any thoughts?

PR #691, opened 2022-06-24
take upstream master without our stuff, and add our anonymous user handling stuff on top

gameguy43:


gameguy43:
whoops, meant to do this in our local copy :P

Issue #695, opened 2022-07-19
AzureAD: Custom signing keys not working due to missing appid


wurstbrot:
### Expected behaviour

Using azuread with customs signed keys should work.

### Actual behaviour

Using azuread with self signed keys the standard `https://login.microsoftonline.com/{tenant}/discovery/v2.0/keys` is loaded and not `https://login.microsoftonline.com/{tenant}/discovery/v2.0/keys?appid={appid}` resulting in _Token error: Cannot find kid=XYZ_.

### Documentation
>If your app has custom signing keys as a result of using the [claims-mapping](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-claims-mapping) feature, you must append an appid query parameter containing the app ID to get a jwks_uri pointing to your app's signing key information, which should be used for validation.

Source: https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens#validating-tokens

### Source code
https://github.com/python-social-auth/social-core/blob/98c81ad7f0a8231c1b4b55fe88064d364c8e9c86/social_core/backends/azuread_tenant.py#L51 and https://github.com/python-social-auth/social-core/blob/98c81ad7f0a8231c1b4b55fe88064d364c8e9c86/social_core/backends/azuread_tenant.py#L49 are interesting here.

I think it would be better to call the discovery endpoint openid-configuration with the appid and extract _jwks_uri_ from the result.

PR #733, opened 2022-12-19
add appid query parameter for azuread

thrstnh:
## Proposed changes

Issue: https://github.com/python-social-auth/social-core/issues/695

Documentation
If your app has custom signing keys as a result of using the claims-mapping ( https://learn.microsoft.com/en-us/azure/active-directory/develop/active-directory-claims-mapping ) feature, you must append an appid query parameter containing the app ID to get a jwks_uri pointing to your app's signing key information, which should be used for validation.
Source: https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens#validating-tokens

## Types of changes

## Checklist

your code._

## Other information

Like @wurstbrot already described in the issue #695, there might be a better fix for this issue, but this seems to work.

### DIFF
--- a/social_core/backends/azuread_tenant.py
+++ b/social_core/backends/azuread_tenant.py
@@ -46,18 +46,25 @@ for verifying JWT tokens.
 
 class AzureADTenantOAuth2(AzureADOAuth2):
     name = "azuread-tenant-oauth2"
-    OPENID_CONFIGURATION_URL = "{base_url}/.well-known/openid-configuration"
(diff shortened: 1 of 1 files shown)


wurstbrot:
I am not having access to an azuread with self created JWKS, but I tried it with a "normal" azuread and it still works (with a patched OWASP DefectDojo instance)

nijel:
Can you please add tests to verify that appid is added to the URL?

nijel:

PR #742, opened 2023-01-27
Allow the OIDC backend to get user data/details from attributes

slyngshede:
Allow the OIDC backend to locate user data in "attributes", if that's where the server puts them.

## Proposed changes

I tried looking of a option that would allow the OIDC backend for get user details from
other places than directly in the response. Having the user details directly in the response
appears to be a legacy feature, at least according to https://apereo.github.io/cas/6.6.x/authentication/OIDC-Authentication-Claims.html.

Instead this patch will use the user data to request the data, but note that it's expect to
be in "attributes" and not as key directly in the response.

I have attempted to preserve the current functionality.

## Types of changes

## Checklist

your code._

## Other information

### DIFF
--- a/social_core/backends/open_id_connect.py
+++ b/social_core/backends/open_id_connect.py
@@ -240,16 +240,33 @@ class OpenIdConnectAuth(BaseOAuth2):
         return response
 
     def user_data(self, access_token, *args, **kwargs):
-        return self.get_json(
+        data = self.get_json(
             self.userinfo_url(), headers={"Authorization": f"Bearer {access_token}"}
         )
 
+        # Some OIDC server implementation with return user data in
+        # attributes.
+        if 'attributes' in data:
+            return data.get('attributes')
+        return data
+
     def get_user_details(self, response):
         username_key = self.setting("USERNAME_KEY", self.USERNAME_KEY)
+
+        # For backwards compatibility, check that username_key
+        # isn't in the response, it only should be for legacy
+        # OIDC.
+        if username_key in response:
+            attrs = response
+        else:
+            # Load user data and get the required user information from
+            # there.
+            attrs = self.user_data(response.get('access_token'))
+
         return {
-            "username": response.get(username_key),
-            "email": response.get("email"),
-            "fullname": response.get("name"),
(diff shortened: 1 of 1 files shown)


davidhalter:
Can you please show where in the OIDC spec these `attributes` are defined? I wasn't able to find it. I also fail to see where in the `Apereo CAS` docs it's mentioned that they have an `attributes` object.

slyngshede:
I don't think it necessarily and OIDC thing and CAS doesn't document it. The response that's returned for the USERINFO_URL looks like this:

```
{
    'access_token': 'xxxxxxxxxxxxxxxxxxxx', 
    'id_token': 'yyyyyyyyyyyyyyyyyyyy',
    'token_type': 'Bearer', 
    'expires_in': 28800, 
    'scope': 'groups email openid profile', 
    'sub': 'Username', 
    'service': 'https://dev-server/complete/oidc/', 
    'auth_time': 1674822029, 
    'attributes': {
        'name': 'User',
        'groups': ['ops', 'dev', 'users']
        'preferred_username': 'slyngshede', 
        'email': 'user@example.org'
    }, 
    'id': 'User', 
    'client_id': 'dev'
}
```

It may very well be CAS that does something wrong. Would it make more sense to create an additional backend, based on the OIDC backend, but which takes this weirdness into account?

slyngshede:
I just notice, that the Patreon backend, while OAuth2, does something similar. 

```
        details = response["attributes"]
        return {
            "username": details.get("full_name"),
            "email": details.get("email"),
            "fullname": details.get("full_name"),
            "first_name": details.get("first_name"),
            "last_name": details.get("last_name"),
        }
```

It's not exactly the same, but it also need to access the attributes for a "sub-dict".

davidhalter:
> It may very well be CAS that does something wrong. Would it make more sense to create an additional backend, based on the OIDC backend, but which takes this weirdness into account?

I'd personally prefer that. IMO we should not complicate the basic OIDC backend, unless there is clearly a general consensus in the "OIDC community" that things should be that way.

slyngshede:
Will rewrite to a separate backend

Issue #749, opened 2023-02-08
IntegrityError at /social-auth/complete/azuread-oauth2/


tiago-peres:
I still get this error

<img width="471" alt="Screenshot 2023-02-08 at 17 44 06" src="https://user-images.githubusercontent.com/13332903/217609985-580f5112-9722-4c3e-bb24-f4fcf5158c3a.png">

which then if I try to sign up with another user leads to

> IntegrityError at /social-auth/complete/azuread-oauth2/
duplicate key value violates unique constraint "social_auth_usersocialauth_provider_uid_e6b5e668_uniq"
DETAIL:  Key (provider, uid)=(azuread-oauth2, None) already exists.

![Screenshot 2023-02-08 at 17 44 34](https://user-images.githubusercontent.com/13332903/217610087-082254f8-cc74-4a76-917a-424cd2b64cf1.png)

_Originally posted by @tiago-peres in https://github.com/python-social-auth/social-core/issues/70#issuecomment-1423007525_

xtsAU:
same issue

PR #750, opened 2023-02-08
fix integrityerror during azure auth

tiago-peres:
**Bug**: https://github.com/python-social-auth/social-core/issues/749

There was someone else solving the exact thing in a different backend: https://github.com/python-social-auth/social-core/pull/298


social_core/backends/azuread.py
```diff
@@ -79,8 +79,8 @@ def access_token_url(self):
         return self.ACCESS_TOKEN_URL.format(base_url=self.base_url)
 
     def get_user_id(self, details, response):
-        """Use upn as unique id"""
-        return response.get("upn")
+        """Use subject (sub) claim as unique id."""
+        return response.get("sub")
```
nijel, line 83: This changes `uid` for existing installation what would be a breaking change. Why `upn` is not present in your claim?

ex8:
I fixed this in my forked codebase. 

https://github.com/ex8/social-core/tree/fix-bug - I have been using that for the time being.

nijel:
This fix is not acceptable to be merged, as it breaks current associations.

nijel:
@ex8 I've created https://github.com/python-social-auth/social-core/pull/1078 out of your changes. Is there any reason why you didn't submit it?

Issue #748, opened 2023-02-08
Authentication with Microsoft Azure AD in a multi-tenant app


tiago-peres:
[According to the Microsoft docs](https://learn.microsoft.com/en-us/troubleshoot/azure/active-directory/error-code-aadsts50020-user-account-identity-provider-does-not-exist?msclkid=ce865af7cc3f11ecb681eb3c54ebc1e1#solution-use-the-correct-sign-in-url),  Multitenant applications are to be redirected to `https://login.microsoftonline.com/organizations`.

Since I'm using `social_core.backends.azuread.AzureADOAuth2`, the user is redirected to `https://login.microsoftonline.com/common` (https://github.com/python-social-auth/social-core/blob/master/social_core/backends/azuread.py#L43). Thing is, `common` is for Multitenant and personal accounts.

I tried using [`social_core.backends.azuread_tenant.AzureADTenantOAuth2`](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/azuread_tenant.py) since it allows to define the target tenant with `SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_TENANT_ID = 'organizations'`, but that didn't work (I reckon is not meant to work that way).

Would be cool to be able to change the Tenant ID in `social_core.backends.azuread.AzureADOAuth2`, like this `SOCIAL_AUTH_AZUREAD_OAUTH2_TENANT_ID = "organizations"`.

[I've explained here in more detail too](https://stackoverflow.com/a/75388348/5675325).

tiago-peres:
For now I created a custom backend extending `AzureADOAuth2`

```
class CustomAzureADOAuth2(AzureADOAuth2):
    BASE_URL = "https://login.microsoftonline.com/organizations"
```

themysteq:
@tiago-peres Strange, it should work but have you got chance to use ` 'social_core.backends.azuread_tenant.AzureADV2TenantOAuth2'`? mind the `V2` version as you should slowly migrate to v2.0 endpoints [See here](https://devblogs.microsoft.com/premier-developer/azure-ad-endpoint-v1-vs-v2/)

PR #751, opened 2023-02-09
add possibility to configure tenant id

tiago-peres:
Bug: https://github.com/python-social-auth/social-core/issues/748

Tested locally and was able to change the Tenant ID to `organizations` with

```
SOCIAL_AUTH_AZUREAD_OAUTH2_TENANT_ID = "organizations"
```

### DIFF
--- a/social_core/backends/azuread.py
+++ b/social_core/backends/azuread.py
@@ -64,7 +64,7 @@ class AzureADOAuth2(BaseOAuth2):
 
     @property
     def tenant_id(self):
-        return "common"
+        return self.setting("TENANT_ID", "common")
 
     @property
     def base_url(self):


nijel:
There already is dedicated backend for tenant authentication: https://github.com/python-social-auth/social-core/blob/master/social_core/backends/azuread_tenant.py

tiago-peres:
@nijel read the bug report.
What I'm changing is not the Tenant, but the assumption that we use `common` in the url all the time.
That's not true and leads to errors doing that.

tiago-peres:
The reason for using is that nomenclature is the way the `BASE_URL` is formed

```
BASE_URL = "https://{authority_host}/{tenant_id}"
```

The `{tenant_id}` is what I'd like have the possibility to change, between `common` or `organizations`, depending on the type of Azure application.

tiago-peres:
can change it to `signin_url` and `SOCIAL_AUTH_AZUREAD_OAUTH2_SIGNIN_URL` if you prefer ([based on Microsoft's documentation](https://learn.microsoft.com/en-us/troubleshoot/azure/active-directory/error-code-aadsts50020-user-account-identity-provider-does-not-exist?msclkid=ce865af7cc3f11ecb681eb3c54ebc1e1#solution-use-the-correct-sign-in-url))

nijel:
I'd like to hear another opinion here as well. Using `TENANT` name for this seems confusing to me - it is not really tenant in this case, but rather a scope.

tiago-peres:
Yea... the reason to use `TENANT` was to make as less changes as possible to what's already there... Look forward to see a resolution though.

nijel:
Instead of misusing `TENANT` I'd rather introduce configuration for sign-in URL, where whole URL would be specified. This way it would be consistent with Azure docs where they document which URL to use. If URL would not be configured, it would fall back to the current implementation.

Issue #753, opened 2023-02-17
Add support for OpenInfra Open ID


ostackbrian:
### Proposed Feature

The OpenId provider at OpenStack was replaced a while ago by OpenInfra and will eventually be decommissioned.  Additionally, there are now users under the OpenInfra umbrella who never had an OpenStackId.  Thus it would be good to add a new backend to support OpenInfraOpenId.

### Any other comments?

I'll be submitting a PR shortly.

nijel:
Fixed via https://github.com/python-social-auth/social-core/pull/754

PR #754, opened 2023-02-17
Add OpenInfraOpenId backend (#753)

ostackbrian:
Add a new backend that's functionally equivalent to the existing OpenStackOpenId backend, but uses the id.openinfra.dev provider.

## Proposed changes

Resolves feature request (issue #753)

## Types of changes

## Checklist

     - social_core.tests.backends.test_dummy.ExpirationTimeTest fails for me locally at line 132; it also fails locally without my patch.  I tried to troubleshoot it, but I'm not sure what's going on there.
     - it's basically the same as an existing backend except for the URL; and the existing backend doesn't have any tests (looks like most of the subclasses of OpenIdAuth don't have unit tests).  I'll be happy to add tests if you can point me at an example.

## Other information

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Backend for OpenInfra OpenID
 - Facebook Limited Login backend
 
 ### Changed
--- a/social_core/backends/openinfra.py
+++ b/social_core/backends/openinfra.py
@@ -1,0 +1,49 @@
+"""
+OpenInfra OpenId backend
+"""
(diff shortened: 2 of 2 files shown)


fungi:
This will be good to have, the (legacy) https://openstackid.org/ in the OpenStackOpenId points users to https://id.openinfra.dev/ for authentication now, as of the middle of last year.

nijel:
The test for OpenID backend is for example https://github.com/python-social-auth/social-core/blob/master/social_core/tests/backends/test_steam.py

Adding documentation to https://github.com/python-social-auth/social-docs/ would be useful as well.

nijel:
@ostackbrian Any chance to look at the tests and documentation?

PR #768, opened 2023-02-26
pre-commit: Replace flake8, isort, and pyupgrade with ruff

cclauss:
## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers
why we should accept this pull request. If it fixes a bug or resolves a feature
request, be sure to link to that issue.

[Ruff](https://beta.ruff.rs) supports [over ~400~ 500 lint rules](https://beta.ruff.rs/docs/rules) including flake8, isort, pylint, and pyupgrade and is written in Rust for speed.

`ruff --format=github .` gives us GitHub Actions annotations:
Ruff annotations

Screenshot 2023-02-26 at 16 44 20
## Types of changes

## Checklist

your code._

## Other information


social_core/backends/azuread_tenant.py
```diff
@@ -2,9 +2,12 @@
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.x509 import load_der_x509_certificate
-from jwt import DecodeError, ExpiredSignatureError
+from jwt import (  # pragma: no cover
```
nijel, line 5: Why the `pragma: no cover` here?

cclauss: Codecov (not my favorite tool) will fail because this line is not covered by a test.

nijel: There are more uncovered lines here, I don't see that a concern. Please remove that comment.

social_core/backends/azuread_tenant.py
```diff
@@ -2,9 +2,12 @@
...
+    DecodeError,
+    ExpiredSignatureError,
+    get_unverified_header,
+)
```
cclauss, line 9:
```suggestion
from jwt import DecodeError, ExpiredSignatureError, get_unverified_header
```

cclauss:
Do we just ignore codecov?

davidhalter:
codecov is just a metric, it is not something that has to get better for all changes. It will also be green again for the next pull request by other people (unlike when you break tests).

.github/workflows/flake8.yml
```diff
@@ -21,7 +21,10 @@ jobs:
     - name: Install pre-commit
       run: |
         python -m pip install --upgrade pip wheel
-        pip install -r requirements-dev.txt
+        pip install ruff -r requirements-dev.txt
```
nijel, line 24: Please avoid installing ruff manually, and especially without a version pin.

.github/workflows/flake8.yml
```diff
@@ -21,7 +21,10 @@ jobs:
...
+
+    - name: Run ruff
+      run: ruff --format github .
```
nijel, line 27: Why not run it via pre-commit?

nijel (commented):
I see you've still kept flake8 in both pre-commit and GitHub workflow. Is there any good reason for that? This way we won't get any performance improvement as both tools would be used.

.github/workflows/flake8.yml
cclauss: Will pre-commit show GitHub annotations to new committers who have not done `pre-commit install`?

cclauss: Oh...  Now I see.  Fixed.

cclauss: Please remove the GitHub Action job `flake8` from required and add `ruff` in its place.

.github/workflows/flake8.yml
cclauss: Done.

.github/workflows/flake8.yml
```diff
@@ -23,8 +23,5 @@ jobs:
         python -m pip install --upgrade pip wheel
         pip install -r requirements-dev.txt
 
-    - name: Run flake8
-      run: |
-        echo "::add-matcher::.github/matchers/flake8.json"
-        pre-commit run flake8 --all-files
-        echo "::remove-matcher owner=flake8::"
```
nijel, line 30: Why remove the matcher? That makes lint issue show up on GitHub.

.github/workflows/flake8.yml
nijel: Only @omab can do such changes, I don't have permissions.

.github/workflows/flake8.yml
cclauss: See the commit message above.  `ruff --format github` makes this _much_ easier.

nijel (approved):
Looks good to me. You can rename the workflow as well – GitHub matches branch protection based on name.

@omab Can you please review this and change branch protection rules to require ruff and pre-commit instead of flake8?

.pre-commit-config.yaml
```diff
@@ -9,24 +9,15 @@ repos:
...
   hooks:
-  - id: isort
-    args: [--profile=black]
-- repo: https://github.com/asottile/pyupgrade
-  rev: v3.3.1
-  hooks:
-  - id: pyupgrade
-    args: [--py36-plus]
+  - id: ruff
+    args: [--format, github]  # --fix, --exit-non-zero-on-fix]
```
nijel, line 16: Maybe args are not needed here when the same is specified in the configuration? Having that specified just once is probably a better approach.

cclauss: If your run `ruff` on the command line or in `pre-commit` then you do not want `--format=github` but if you run it in GitHub Actions then you do.  So, put the majority of config in `pyproject.toml` and only `--format` in args.

nijel: But this does not address only running inside github action, this is generic pre-commit configuration...

pyproject.toml
```diff
@@ -0,0 +1,21 @@
...
+  ".venv",
+  "*.egg",
+  "build",
+  "db/env.py",
+  "db/versions/*.py",
+  "dist",
+  "doc",
+  "site"
+]
+format = "github"
```
nijel, line 14: ...and here the github format is specified as well.

cclauss:
Now I see.  Fixed.  I have provided two different versions of ruff action.  One runs in < 10 sec for fast feedback to contributors.  The other goes thru pre-commit and is much slower.

pyproject.toml
cclauss: Removed.

pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+  "build",
+  "db/env.py",
+  "db/versions/*.py",
+  "dist",
+  "doc",
+  "site"
+]
+line-length = 120
+select = ["E", "F", "I", "PLC", "PLE", "UP"]
+target-version = "py37"
```
nijel, line 16: As GitHub format is not really something what would be annoying to use, I think we can stick with it being default here as well.

cclauss: Did you try running ruff on the command line with `--format=github` and some errors?

PR #459, opened 2020-05-07
Update Azure AD B2C base url to the latest url as stated in the Microsoft docs

jonatns:
## Proposed changes

Using login.microsoftonline.com for Azure AD B2C has been deprecated and it will be removed later this year. See [Deprecation notice](https://docs.microsoft.com/en-us/azure/active-directory-b2c/b2clogin). Even though login.microsoftonline.com is still being supported by Microsoft I was not able to get it working for a new Azure AD B2C Tenant. Updating the BASE_URL to the new url and moving the policy to the url itself instead of a query parameter works correctly. This is how it's stated in the latest [docs for Azure AD B2C OpenID Connect](https://docs.microsoft.com/en-us/azure/active-directory-b2c/openid-connect).

## Types of changes

Please check the type of change your PR introduces:

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (PEP8, lint, formatting, renaming, etc)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build related changes (build process, tests runner, etc)
- [ ] Other (please describe):

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask. We're here to
help! This is simply a reminder of what we are going to look for before merging
your code._

- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works

## Other information

SOCIAL_AUTH_AZUREAD_B2C_OAUTH2_TENANT_ID was changed to SOCIAL_AUTH_AZUREAD_B2C_OAUTH2_TENANT_NAME since only the tenant name is needed now.

PR #777, opened 2023-03-21
Update Azure AD B2C base URL to match updated endpoints

zchoate:
## Proposed changes

This is similar to [PR 459](https://github.com/python-social-auth/social-core/pull/459). I kept the policy name as an extra argument. This appears to be the default endpoint suggested by Azure AD B2C which is probably less confusing. I also made the b2clogin.com endpoint using the `authority_host` property. I have updated the corresponding tests as well as the docker-compose.yml to test with these Python versions: 3.6.12 3.7.16 3.8.16 3.9.16 3.10.10 3.11.2. I have documentation updates drafted and I can propose those changes if this PR is accepted. 

Documentation for the endpoints can be found here: https://docs.microsoft.com/en-us/azure/active-directory-b2c/b2clogin

As proposed the `base_url` would be `https://{tenant_name}.{authority_host}/{tenant_name}.onmicrosoft.com` where `tenant_name` is required but `authority_host` will default to `b2clogin.com` if not set. The policy argument will still get added so an example URL for keys would look like this `https://footenant.b2clogin.com/footenant.onmicrosoft.com/discovery/v2.0/keys?p=b2c_1_signin`.

## Types of changes

## Checklist

your code._

## Other information

`SOCIAL_AUTH_AZUREAD_B2C_OAUTH2_TENANT_ID` is updated to `SOCIAL_AUTH_AZUREAD_B2C_OAUTH2_TENANT_NAME`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,8 +19,8 @@ services:
       context: .
       dockerfile: ./files/tests/Dockerfile
       args:
-      - PYTHON_VERSIONS=3.6.12 3.7.9 3.8.7 3.9.1
+      - PYTHON_VERSIONS=3.6.12 3.7.16 3.8.16 3.9.16 3.10.10 3.11.2
     environment:
-    - PYTHON_VERSIONS=3.6.12 3.7.9 3.8.7 3.9.1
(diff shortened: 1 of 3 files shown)


nijel:

Issue #538, opened 2021-01-07
[CRITICAL] Steam OpenID implementation allows spoofing


bommels:
I am using the latest version of social-core and social-app-django.

### Expected behaviour

Steam OpenID redirects from a different server than 'https://steamcommunity.com/openid/login' should not get authenticated.

### Actual behaviour

Spoofing the OpenID server (by for example, hosting your own), and redirecting to the /complete/ endpoint authenticates the user. There is no check on claimed_id, identity and op_endpoint.

### What are the steps to reproduce this issue?

1. Try to auth through Steam
2. Authenticate with your Steam account, capture the redirect url (.../complete/steam...)
3. Replace steamcommunity.com in the URL with your own OpenID host
4. You will be authenticated

You can let your own OpenID instance return any steam id you want to authenticate as that steam id/user. I have tested this, and could successfully authenticate as any steam user.

### Any other comments?

I fixed this issue locally by implementing the following and adding it to the auth pipeline, I am not 100% sure if this covers the bug fully.

```python
def steam_auth_check(*args, **kwargs):
    if kwargs['response'].message.args[('http://specs.openid.net/auth/2.0', 'claimed_id')][0:27] != 'https://steamcommunity.com/':
        raise Exception('Authentication Mismatch')
    if kwargs['response'].message.args[('http://specs.openid.net/auth/2.0', 'identity')][0:27] != 'https://steamcommunity.com/':
        raise Exception('Authentication Mismatch')
    if kwargs['response'].message.args[('http://specs.openid.net/auth/2.0', 'op_endpoint')] != 'https://steamcommunity.com/openid/login':
        raise Exception('Authentication Mismatch')
```

nijel:
The Steam backend was intentionally made stateless in d97584134, maybe that no longer applies these days. Can you try removing the `consumer` method from `SteamOpenId` and chech whether it works well then?

PR #807, opened 2023-07-08
fix: SteamOpenId does not validate identity url

zhuang42:
## Proposed changes

The implementation of SteamOpenId does not validate identity URL and could be impersonated by attackers.

## Types of changes

## Checklist

your code._

## Other information

### DIFF
--- a/social_core/backends/steam.py
+++ b/social_core/backends/steam.py
@@ -45,7 +45,7 @@ class SteamOpenId(OpenIdAuth):
         return self._consumer
 
     def _user_id(self, response):
-        if not response.identity_url.startswith(SteamOpenId.URL):
+        if not response.identity_url.startswith(self.URL):
             raise AuthFailed(self, "Openid identifier mismatch")
         user_id = response.identity_url.rsplit("/", 1)[-1]
         if not user_id.isdigit():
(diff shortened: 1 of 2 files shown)


zhuang42:
#538

nijel:
Anyone tried my suggestion in https://github.com/python-social-auth/social-core/issues/538#issuecomment-787514783 instead? This patch IMHO doesn't address https://github.com/python-social-auth/social-core/issues/538 at all.

zhuang42:
The Steam openId login can still be stateless. Also I don't see how adding stores in https://github.com/python-social-auth/social-core/issues/538#issuecomment-787514783 address the problem, since it doesn't validate the op_endpoint/claim_id from the response still. Could you explain more about your fix?

nijel:
If I understood https://github.com/python-social-auth/social-core/issues/538 correctly, the identify URL can be spoofed by third-party server. Restricting it to spoofing valid URL doesn't IMHO address the issue. What we really need is validating that the assertion is coming from the correct server, and keeping state should ensure that. (I might be wrong, my OpenID knowledge is very limited)

social_core/backends/steam.py
```diff
@@ -45,6 +45,8 @@ def consumer(self):
         return self._consumer
 
     def _user_id(self, response):
+        if not response.identity_url.startswith(SteamOpenId.URL):
```
nijel, line 48:
```suggestion
        if not response.identity_url.startswith(self.URL):
```

nijel:
Merged as this does not harm, but I'm still not convinced this really addresses the original issue.

PR #815, opened 2023-07-21
Make Keycloak's ID_KEY configurable

### LINKED ISSUE #813: Keycloak ID_KEY should be read from settings
The docstring in [social_core/backends/keycloak.py#L88-L91](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/keycloak.py#L88-L91) states:

> 8. The default behaviour is to associate users via username field, but you
>       can change the key with e.g.
>       ```
>        SOCIAL_AUTH_KEYCLOAK_ID_KEY = 'email'
>       ```

However, the `ID_KEY` is hard-coded instead of being read from the settings, see [social_core/backends/keycloak.py#L153](https://github.com/python-social-auth/social-core/blob/master/social_core/backends/keycloak.py#L153).

derlin:
## Proposed changes

As explained in:
- #813 

Currently, it is not possible to change the `ID_KEY` of the keycloak provider, even though this feature was meant to be present (according to the docstring). The `ID_KEY` is basically the claim used to map a user.

With this change, it is now possible to set the `ID_KEY` to any field present either in details or in the returned token:
```python
SOCIAL_AUTH_KEYCLOAK_ID_KEY = "email" # use the keycloak user email (from details)
SOCIAL_AUTH_KEYCLOAK_ID_KEY = "sub" # use the keycloak user id (from token)
```

## Types of changes

## Checklist

your code._

## Other information

I haven't added tests since such configurations don't seem to be tested on other backends. The default will stay exactly the same, thus not breaking the current behaviour.

Closes #813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Fixed Azure AD Tenant authentication with custom signing keys
 - Added CAS OIDC backend
+- Made Keycloak `ID_KEY` configurable
 
 ## [4.4.1](https://github.com/python-social-auth/social-core/releases/tag/4.4.1) - 2023-03-30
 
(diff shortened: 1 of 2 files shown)


nijel:

PR #832, opened 2023-09-22
make request token method Trelll compatible

bs-bhagirath-radadiya:
if social auth for Trello then scope arguments are not getting for the request token method

### DIFF
--- a/social_core/backends/oauth.py
+++ b/social_core/backends/oauth.py
@@ -247,7 +247,8 @@ class BaseOAuth1(OAuthAuth):
     def unauthorized_token(self):
         """Return request for unauthorized token (first stage)"""
         params = self.request_token_extra_arguments()
-        # params.update(self.get_scope_argument())
+        if self.name not in ['trello']:
+            params.update(self.get_scope_argument())
         key, secret = self.get_key_and_secret()
         state = self.get_or_create_state()
         response = self.request(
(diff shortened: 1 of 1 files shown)


social_core/backends/oauth.py
```diff
@@ -247,7 +247,8 @@ def request_token_extra_arguments(self):
     def unauthorized_token(self):
         """Return request for unauthorized token (first stage)"""
         params = self.request_token_extra_arguments()
-        params.update(self.get_scope_argument())
+        if self.name not in ['trello']:
+            params.update(self.get_scope_argument())
```
nijel, line 251: If this is really needed, overriding `get_scope_argument` in `TrelloOAuth` is the way to go.

nijel (changes requested):
Can you please describe what issue are you trying to address?

social_core/backends/oauth.py
bhagirathr: Trello REQUEST_TOKEN_URL (https://trello.com/1/OAuthGetRequestToken) does not take any arguments.

If we send the scope here then the URL looks like this **https://trello.com/1/OAuthGetRequestToken?scope=read,write**

As per above mentioned Trello request does not take any arguments. Trello gives an error like this...

ERROR: 400 Client Error: Bad Request for URL: https://trello.com/1/OAuthGetRequestToken?scope=read%2Cwrite

nijel: Thanks, this should go as a comment to the code. And the implementation should override `get_scope_argument` method in `TrelloOAuth` instead of doing changes in generic OAuth backend.

nijel:
Closing as review feedback was not addressed.

PR #851, opened 2023-11-07
Pass "next" URL through SAML RelayState

RJPercival:
Passing the "next" URL to redirect to after authentication completes is the primary use case for `RelayState`, but the SAML backend did not previously support this use. Instead, it relied on the `Strategy` to store the "next" URL in session state and restore it from there after login. Unfortunately, in the case of SAML, this does not work if the server has set the session cookie's [`SameSite` attribute](https://owasp.org/www-community/SameSite) to "Lax" or "Strict", as [the SAML POST request from the IdP](https://en.wikipedia.org/wiki/SAML_2.0#SP_redirect_request;_IdP_POST_response) will not contain the session cookie.

The new `RelayState` format (a JSON object) allows for further extensibility, as arbitrary additional fields can be added in the future.

A number of additional SAML response test fixtures have been added - these differ only in the value of the `RelayState` parameter, and are used for testing different values (e.g. missing keys in the JSON).

I've manually tested this with Okta and Django and the redirect works as intended (it previously did not).


digismack (approved):

digismack:
Looks good to me. Thanks!

RJPercival:
Looks like I'm not authorized to merge this @digismack.

RJPercival:
@digismack, thanks for merging. Any idea when this will go out in a release?

nijel:
We need a new release soon to address https://github.com/python-social-auth/social-core/issues/854, but it needs to be fixed first…

PR #819, opened 2023-08-03
Replace jose with pyjwt

sevdog:
## Proposed changes

Replace `python-jose` with `pyjwt` (see #503).

## Types of changes

Please check the type of change your PR introduces:

- [ ] Release (new release request)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (PEP8, lint, formatting, renaming, etc)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build related changes (build process, tests runner, etc)
- [ ] Other (please describe):

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask. We're here to
help! This is simply a reminder of what we are going to look for before merging
your code._

- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works

## Other information

As discussed in #503 some changes was done to perform the migration:

- add `calc_at_hash` static method to `OpenIdConnectAuth` since PyJWT does not support validating claims which are not part of JWT RFCs (see https://github.com/jpadilla/pyjwt/pull/296)
- manual add `"at_hash"` claim in `OpenIdConnectTestMixin.prepare_access_token_body` (see above)
- add `"aud"` claim in `Auth0OAuth2Test.access_token_body` because PyJWT behaves different from jose: it does not allow this claim to not be present if is should be validated
- implement in `Auth0OAuth2` the logic which tries every found key since PyJWT does not implement it while jose does.
- update PyJWT minimum requirement to `2.7.0` which is the minimum version which provides the full API to handle OIDC

PR #852, opened 2023-11-07
Skip 'at_hash' claim validation when missing

### LINKED ISSUE #849: OpenID Connect "at_hash" claim requirement during ID token validation
### Expected behaviour

Per the [OpenID Connect specification](https://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken), the `at_hash` claim in ID Token _can_ be optional, depending on circumstance. Thus, I'd expect ID token validation to succeed when the `at_hash` ID cliam is not required and not provided.

This was the actual behavior up to social-core 4.4.2.

### Actual behaviour

Since social-core version 4.5, the `OpenIdConnectAuth` backend incorrectly raises an `AuthTokenError` when the Authorization Code flow is used without the use of the `at_hash` claim in an ID token. 

Input clear steps to reproduce the issue for a maintainer.

1. Use the `OpenIdConnectAuth` backend with the `code` `RESPONSE_TYPE`.
2. Make a request to a ID provider which does not include an `at_hash` claim in the ID token, such as Auth0.com.
3. Error occurs upon ID token validation.

The problematic code in question is [here](https://github.com/python-social-auth/social-core/blob/6c69c603a6d2e614754ad3cb3e8de55ab46d159d/social_core/backends/open_id_connect.py#L239). This check should succeed when `at_hash` is not in the claim set, as is the behavior in the original `python-jose` [functionality](https://github.com/mpdavis/python-jose/blob/4b0701b46a8d00988afcc5168c2b3a1fd60d15d8/jose/jwt.py#L444C11-L444C11).

### Any other comments?

browniebroke:
## Proposed changes

Fix validation error when the `at_hash` claim is absent. 
- Fix #849
- Refs #849
- Refs #819 

## Types of changes

## Checklist

your code._

## Other information

### DIFF
--- a/social_core/backends/open_id_connect.py
+++ b/social_core/backends/open_id_connect.py
@@ -236,7 +236,9 @@ class OpenIdConnectAuth(BaseOAuth2):
 
         # pyjwt does not validate OIDC claims
         # see https://github.com/jpadilla/pyjwt/pull/296
-        if claims.get("at_hash") != self.calc_at_hash(access_token, key["alg"]):
+        if "at_hash" in claims and claims["at_hash"] != self.calc_at_hash(
+            access_token, key["alg"]
+        ):
             raise AuthTokenError(self, "Invalid access token")
 
         self.validate_claims(claims)


social_core/backends/open_id_connect.py
```diff
@@ -236,7 +236,9 @@ def validate_and_return_id_token(self, id_token, access_token):
 
         # pyjwt does not validate OIDC claims
         # see https://github.com/jpadilla/pyjwt/pull/296
-        if claims.get("at_hash") != self.calc_at_hash(access_token, key["alg"]):
+        if "at_hash" in claims and claims["at_hash"] != self.calc_at_hash(
+            access_token, key["alg"]
+        ):
             raise AuthTokenError(self, "Invalid access token")
```
browniebroke, line 242: As far as I can tell, this isn't covered by the tests currently.

nijel:

social_core/backends/open_id_connect.py
nijel: Apparently, there is no test where at_hash claim would be missing, so this behavior is not tested (otherwise we would see a test failure earlier).

Issue #879, opened 2024-01-30
bugs: Etsy OAuth2 Authentication


amitray007:
### Expected behaviour

OAuth2 login should work normally.

### Actual behaviour
- Exception as username should be a string but Etsy returns integer `user_id`
- PKCE Method from Etsy should S256 and not s256 as per `BaseOAuth2PKCE`
- REDIRECT_STATE should not be added to redirect URI.

### What are the steps to reproduce this issue?

Input clear steps to reproduce the issue for a maintainer.

1. Just try logging in, It'll fail.

PR #880, opened 2024-01-30
fix: Etsy OAuth2 Authentication bugs

amitray007:
## Proposed changes

Resolves Issue: https://github.com/python-social-auth/social-core/issues/879

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [X] I have added documentation to https://github.com/python-social-auth/social-docs

## Other information

Note: This is a breaking change and could potentially break authentications.

### DIFF
--- a/social_core/backends/oauth.py
+++ b/social_core/backends/oauth.py
@@ -494,7 +494,7 @@ class BaseOAuth2PKCE(BaseOAuth2):
 
     def generate_code_challenge(self, code_verifier, challenge_method):
         method = challenge_method.lower()
-        if method in "s256":
+        if method == "s256":
             hashed = hashlib.sha256(code_verifier.encode()).digest()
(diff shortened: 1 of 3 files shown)


social_core/backends/oauth.py
```diff
@@ -494,7 +494,7 @@ def get_code_verifier(self):
 
     def generate_code_challenge(self, code_verifier, challenge_method):
         method = challenge_method.lower()
-        if method == "s256":
+        if method in ["s256", "S256"]:
```
mmd-osm, line 497: „method“ should be lowercase here due to the previous line 496. Why would this change be needed?

amitray007: You're right. Removed this change.

nijel:

Issue #109, opened 2017-06-30
token validation for azure AD backend


sdementen:
Does the azure AD Backend validates the id_token and access_token sent by Azure AD?

In the code I see a `decoded_id_token = jwt_decode(id_token, verify=False)`. 

When I set to `verify=True`:

- I get an exception "Token error: Signature verification failed" if I tamper with the id_token which is fine
- I get an exception "Could not deserialize key data." if the id_token is real which is not OK

My questions:
- shouldn't the "verify=True" be the default ?
- how can I fix the "Could not deserialize key data" exception ?

bj00rn:
I'm facing the same issue. 

After a bit of research it seems that the Azure AD tokens are signed by public key and the key needs to fetched from provided to the decode function according to the [docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens#validating-the-signature)

> Tokens issued by Azure AD are signed using industry standard asymmetric encryption algorithms, such as RSA 256. The header of the JWT contains information about the key and encryption method used to sign the token.

> The alg claim indicates the algorithm that was used to sign the token, while the kid claim indicates the particular public key that was used to validate the token.

> You can acquire the signing key data necessary to validate the signature by using the OpenID Connect metadata document located at: https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration 

> The JSON document located at the jwks_uri contains all of the public key information in use at that particular moment in time. Your app can use the kid claim in the JWT header to select which public key in this document has been used to sign a particular token. It can then do signature validation using the correct public key and the indicated algorithm.

To make matters worse:

> At any given point in time, Azure AD may sign an id_token using any one of a certain set of public-private key pairs. Azure AD rotates the possible set of keys on a periodic basis, so your app should be written to handle those key changes automatically. A reasonable frequency to check for updates to the public keys used by Azure AD is every 24 hours.

bj00rn:
Here are the relevant parts from an npm [package](https://github.com/AzureAD/passport-azure-ad/) that would need to be implemented

https://github.com/AzureAD/passport-azure-ad/blob/fcf9b755a1bbe7a70c7340e2fdd36e8813aa8546/lib/metadata.js#L69

https://github.com/AzureAD/passport-azure-ad/blob/fcf9b755a1bbe7a70c7340e2fdd36e8813aa8546/lib/metadata.js#L96

[...]

PR #910, opened 2024-05-03
[Security fix] AzureADOAuth2 backend

mike667:
## Proposed changes

Hello, my changes involve replacing the user_data method for AzureADOAuth2. Since the old method simply decoded data from the JWT token without verifying its signature, any user could modify data in the JWT and authenticate on the site. In the new method, the data is extracted from the Microsoft Graph API using the access token, which fixes this vulnerability. Additionally, aliases for data have been added so that in future pipelines, functions like associate_by_email and user_details can retrieve the data.

Please tell me if these changes make sense, if so, I will update the tests. (Currently, this code is in production on my project.)

## Types of changes

## Checklist

your code._

## Other information

### DIFF
--- a/social_core/backends/azuread.py
+++ b/social_core/backends/azuread.py
@@ -1,6 +1,5 @@
 import time
 
-from ..exceptions import AuthTokenError
 from .oauth import BaseOAuth2
 
 """
--- a/social_core/backends/azuread.py
+++ b/social_core/backends/azuread.py
@@ -80,35 +80,34 @@ class AzureADOAuth2(BaseOAuth2):
         return response.get("upn")
 
     def get_user_details(self, response):
-        """Return user details from Azure AD account"""
-        fullname, first_name, last_name = (
-            response.get("name", ""),
-            response.get("given_name", ""),
-            response.get("family_name", ""),
-        )
+        """Return user details from Microsoft online account"""
+        email = response.get("mail")
+        username = response.get("userPrincipalName")
+
+        if "@" in username:
+            if not email:
+                email = username
+            username = username.split("@", 1)[0]
+
         return {
-            "username": fullname,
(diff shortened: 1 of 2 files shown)


mike667:
Related to [Issue](https://github.com/python-social-auth/social-core/issues/109)

social_core/backends/azuread.py
```diff
@@ -83,32 +80,34 @@ def get_user_id(self, details, response):
...
-            response.get("family_name", ""),
-        )
+        """Return user details from Microsoft online account"""
+        email = response.get("mail")
+        username = response.get("userPrincipalName")
+
+        if "@" in username:
+            if not email:
+                email = username
+            username = username.split("@", 1)[0]
```
nijel, line 90: `username` was based on `name` before.

social_core/backends/azuread.py
```diff
@@ -83,32 +80,34 @@ def get_user_id(self, details, response):
...
+
         return {
-            "username": fullname,
-            "email": response.get("email", response.get("upn")),
-            "fullname": fullname,
-            "first_name": first_name,
-            "last_name": last_name,
+            "username": username,
+            "email": email,
```
nijel, line 94: `email` used to fall back to `userPrincipalName` if not present, this is now gone.

nijel (commented):
I think this change makes sense as this is similar what other OAuth2 backends do. Do you see any risks in breaking existing users? I think some details will have a different value after your changes.

mike667:
Hi, after thinking about it a bit, there is no point in changing AzureADOAuth2 if it is very similar to MicrosoftOAuth2. It is better to just use MicrosoftOAuth2, and add a note for AzureADOAuth2 so that new users understand that AzureADOAuth2.user_data is not completely secure. Therefore, I am closing this request pool and opening a new one for documentation.

Issue #602, opened 2021-07-09
ORCID: backend crashes when no family name set in ORCID account


byte-for-byte:
### Expected behaviour

Successful login through ORCID backend.

### Actual behaviour

Authentication crashes with:
AttributeError: 'NoneType' object has no attribute 'get'

### What are the steps to reproduce this issue?

1) Create an ORCID account WITHOUT specifying a "Last Name".
2) Try social auth

### Any logs, error output, etc?

if "Last Name" is not provided in the ORCID account, 
name in .../backends/orcid.py", about line 77 will be 
name = {'created-date': {.... 'family-name': None, ....}

and in line 77, last_name = name.get('family-name', {}).get('value', '')

a get is used on a None type.

File ".../lib/python3.9/site-packages/social_core/backends/orcid.py", line 77, in get_user_details
  last_name = name.get('family-name', {}).get('value', '')
AttributeError: 'NoneType' object has no attribute 'get'

nikoder:
This should be resolved on `master`, as of commit `95316b0914386af59d8649917c373a594f1766de`.

PR #914, opened 2024-05-15
Fix ORCID login when no family name was given

nikoder:
Handle the case where the user has not registered a `family-name` and ORCID returns `None`.

## Proposed changes

This addresses [https://github.com/python-social-auth/social-app-django/issues/355](https://github.com/python-social-auth/social-app-django/issues/355) and #602 (which it appears to duplicate). Full credit to [paloha](https://github.com/paloha), who very clearly [explained](https://github.com/python-social-auth/social-app-django/issues/355#issuecomment-1568685404) the issue and proposed the simple fix included here.

As a brief summary: ORCID does not require users to enter a family name and returns a `family-name` key with value `None` in this case. The code expects that key to either be absent, or contain a dictionary. This leads to an `AttributeError` after successful login by a user who has not provided a family name to ORCID.

I considered adding a test case to cover this scenario, but it looks like the testing infrastructure is only set up for a single possible `user_data_body`? Please let me know if I have missed an easy and obvious way to add a test for an additional case of returned data. The change does resolve the issue in my manual testing.

## Types of changes

## Checklist

your code._

- [x] Lint and unit tests pass locally with my changes - caveat: 6 of 587 discovered tests fail independently of these changes in my environment but they do not appear to be covering the changed code (they are related to SAML)

## Other information

_Any other information that is important to this PR such as screenshots of how
the component looks before and after the change._

### DIFF
--- a/social_core/backends/orcid.py
+++ b/social_core/backends/orcid.py
@@ -76,7 +76,8 @@ class ORCIDOAuth2(BaseOAuth2):
 
             if name:
                 first_name = name.get("given-names", {}).get("value", "")
-                last_name = name.get("family-name", {}).get("value", "")
+                if name.get("family-name", None) is not None:
+                    last_name = name.get("family-name").get("value", "")
 
             emails = person.get("emails")
             if emails:
--- a/social_core/backends/orcid.py
+++ b/social_core/backends/orcid.py
@@ -76,8 +76,8 @@ class ORCIDOAuth2(BaseOAuth2):
 
             if name:
                 first_name = name.get("given-names", {}).get("value", "")
-                if name.get("family-name", None) is not None:
-                    last_name = name.get("family-name").get("value", "")
+                if (family_name := name.get("family-name", None)) is not None:
+                    last_name = family_name.get("value", "")
 
             emails = person.get("emails")
             if emails:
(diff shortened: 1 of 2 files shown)


social_core/backends/orcid.py
```diff
@@ -76,7 +76,8 @@ def get_user_details(self, response):
 
             if name:
                 first_name = name.get("given-names", {}).get("value", "")
-                last_name = name.get("family-name", {}).get("value", "")
+                if name.get("family-name", None) is not None:
+                    last_name = name.get("family-name").get("value", "")
```
nijel, line 80:
```suggestion
                if (family_name := name.get("family-name", None)) is not None:
                    last_name = family_name.get("value", "")
```

nikoder: Sounds good, thanks!

I'll replace the existing commit with that change for a simpler commit history - hope that is what you would prefer to (let me know if not, I'm happy to change it back to two separate commits if that is your preference :) ) .

nikoder: Oh, actually, I think there is a feature to squash commits on merge in Github, perhaps? I'll just make it 2 commits for now, for easier review - still happy to squash it externally, of course (just let me know if you would like me to do so :) ) .

nikoder: Just confirming that all unit tests and my manual tests still pass the same as before with this change :+1:

nijel:

nikoder:
My pleasure! Thanks for your great work on this awesome project :) !

PR #926, opened 2024-07-17
Use timezone-aware datetime objects

Tenzer:
Rather than using the `datetime.utcnow()` and `datetime.utcfromtimestamp()` methods which produce timezone-naive datetime objects, we should instead use `datetime.now()` and `datetime.fromtimestamp()` with UTC provided as the timezone, in order to create timezone-aware datetime objects.

This removes the DeprecationWarning that otherwise is raised since Python 3.12 for these methods.

## Types of changes

## Checklist


nijel:

PR #927, opened 2024-07-24
Fix failing Elixir AAI token refresh

### LINKED ISSUE #826: OAuth2 grant type `refresh_token` sends the access token when no refresh token is available
### Expected behaviour

According to [RFC6749 section 4.1.4](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.4), when requesting an authorization code via OAuth2 (`grant_type=authorization_code`) using a request like this,

```
POST /token HTTP/1.1
Host: server.example.com
Authorization: Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW
Content-Type: application/x-www-form-urlencoded

grant_type=authorization_code&code=SplxlOBeZQQYbYS6WxSbIA
&redirect_uri=https%3A%2F%2Fclient%2Eexample%2Ecom%2Fcb
```

the response _may_ contain a `refresh_token`.

```
HTTP/1.1 200 OK
Content-Type: application/json;charset=UTF-8
Cache-Control: no-store
Pragma: no-cache

{
    "access_token": "2YotnFZFEjr1zCsicMWpAA",
    "token_type": "example",
    "expires_in": 3600,
    "refresh_token": "tGzv3JOkF0XG5Qx2TlKWIA",
    "example_parameter": "example_value"
}
```

If the response contains no `refresh_token` and later [`refresh_token()`](https://github.com/python-social-auth/social-core/blob/4a7d470dee9cda3ffaf2d32f6f275d466b6f12a1/social_core/storage.py#L46-L55) is called, I would expect `social-core` to raise an exception.

### Actual behaviour

`social-core` [_sends the access token_ as if it was a refresh token](https://github.com/python-social-auth/social-core/blob/4a7d470dee9cda3ffaf2d32f6f275d466b6f12a1/social_core/storage.py#L47-L49) and the authorization server reports that the refresh token is invalid.

Request an authorization code via OAuth2 to an authorization server that does not return a `refresh_token` (for example, [LS Login](https://lifescience-ri.eu/ls-login/)).

The authorization server returns a response with error code 401.

```
An error occurred when refreshing user token: 401 Client Error: 401 for url: https://login.elixir-czech.org/oidc/token
```

The error is arising in version [4.5.4](https://github.com/python-social-auth/social-core/releases/tag/4.5.4).

### Any other comments?

We have spotted this issue on [UseGalaxy.eu](https://usegalaxy.eu/), that makes use of LS Login in production. This problem arises about one hundred thousand times a day since September 1, 2023.

kysrpex:
## Proposed changes

Refreshing Elixir AAI tokens fails with error `An error occurred when refreshing user token: 401 Client Error: 401 for url: https://login.elixir-czech.org/oidc/token`. The source of the HTTP 401 error is a malformed request. Closes #826, read the issue for details.

## Types of changes

## Checklist

your code._

## Other information

At the moment I just want to highlight the source of the problem and the solution, the patch is not properly integrated within the codebase, that's why the PR is a draft. For example, it is probably better to patch [`social_core.backends.oauth:BaseOAuth2.refresh_token`](https://github.com/python-social-auth/social-core/blob/fd7dd903eedd666027ffee65bf8f5f39647e9fbe/social_core/backends/oauth.py#L454-L461).

### DIFF
--- a/social_core/backends/elixir.py
+++ b/social_core/backends/elixir.py
@@ -32,3 +32,14 @@ class ElixirOpenIdConnect(OpenIdConnectAuth):
             "first_name": first_name,
             "last_name": last_name,
         }
+
+    def refresh_token(self, token, *args, **kwargs):
+        params = "&".join(
+            f"{param}={value}"
+            for param, value in self.refresh_token_params(token, *args, **kwargs).items()
+        )
+	    url = self.refresh_token_url()
+        method = self.REFRESH_TOKEN_METHOD
+        key = "params" if method == "GET" else "data"
+        request_args = {"headers": self.auth_headers(), "method": method, key: params}
+        return self.process_refresh_token_response(request, *args, **kwargs)


dBucik:
From the OP point of view, the authentication method for the token endpoint should be configurable and the refresh token grant should respect that setting.
Specifically for the LS AAI (previously known as ELIXIR AAI), there are three types of authentication for the token endpoint:

- client_secret_basic (translates to `Authorization: Basic client_id:client_secret` header)
- client_Secret_post (translates to `client_id: XYZ` and `client_secret: XYZ` parameters in the POST request body)
- none (no authentication needed)

This method of authentication is specified in the metadata when registering the service.

PR #931, opened 2024-08-22
Add AzureADOauth2 backend using the v2.0 API.

### LINKED ISSUE #723: Azure Active Directory is obsolete and broken.
### Expected behaviour

The flow goes through

### Actual behaviour

Authentication page returns an error

Create an azure active directory app registration, and try to use it.

### Any other comments?

This problem is due to the fact that the default scopes requested by social core, https://github.com/python-social-auth/social-core/blob/8247ee47fb6d79ddb0196e8b7b8a532da5b02bbb/social_core/backends/azuread.py#L48, `['openid', 'profile', 'user_impersonation', 'email']` which are azure AD graph permissions, are deprecated and no longer usable since june 2022.

I fixed it by replacing these scopes with the new microsoft graph : `['user.read']`. 

This isn't the only issue, i also faced a really strange issue when the lib attempts to decode the JWT returned. They ... do not contain a signature ?
This issue seems exclusive to tenant auth. It happens exactly here : https://github.com/python-social-auth/social-core/blob/8247ee47fb6d79ddb0196e8b7b8a532da5b02bbb/social_core/backends/azuread_tenant.py#L88, `KeyError: 'kid'`. Here is what is returned by `get_unverified_header` : `{'typ': 'JWT', 'alg': 'none'}`, and when printing the `id_token` it seems to simply not hold a signature. I cannot show it since it contains personal data, but it has the header and the body separated by a period, and another trailing period after the body.

mateuszmandera:
`AzureADOAuth2` uses the v1.0 API which doesn't support personal accounts. Updating the endpoints used by the original class may break backward compatibility, so add this as just an additional subclass.

This should also fix #723

### DIFF
--- a/social_core/backends/azuread.py
+++ b/social_core/backends/azuread.py
@@ -145,3 +145,14 @@ class AzureADOAuth2(BaseOAuth2):
             new_token_response = self.refresh_token(token=access_token)
             access_token = new_token_response["access_token"]
         return access_token
+
+
+class AzureADOAuth2V2(AzureADOAuth2):
+    """Version of the AzureADOAuth2 backend that uses the v2.0 API endpoints,
+    supporting users with personal Microsoft accounts, if the app settings
+    allow them."""
+
+    name = "azuread-oauth2-v2"
+    AUTHORIZATION_URL = "{base_url}/oauth2/v2.0/authorize"
+    ACCESS_TOKEN_URL = "{base_url}/oauth2/v2.0/token"
+    DEFAULT_SCOPE = ["User.Read profile openid email"]


nijel:
Is it needed to have two backends? Maybe just the current backend could be upgraded?

mateuszmandera:
@nijel I wasn't sure if there isn't some incompatible behavior between these APIs that some old code using this backend might rely on. At least there are differences in supported scopes, e.g. in the new API you want to have the `User.Read` scope enabled; I believe that wasn't supported in `v1.0`.  So if a project overrode the scopes when subclassing `AzureADOAuth2`, swapping the endpoint will break things for them upon upgrading `social-core`.

nijel:
Okay, if this needs a configuration change, then a separate backend is probably the way to go. I was just wondering if this could be silently upgraded so that once 1.0 API is disabled, the backend will continue to work.

mateuszmandera:
I haven't found any information about Microsoft having a concrete plan to disable the 1.0 API, so it might also very well be that the old backend will keep working for a long time. So in my view not worth it to risk giving users trouble by possibly breaking their integration with backward incompatible behavior by swapping the endpoint.

nijel:

PR #933, opened 2024-08-30
Fix access token expiration and refresh token handling in GitHub backend

### LINKED ISSUE #932: Fix access token expiration and refresh token handling in GitHub backend
The `access_token` expiration value is not stored correctly by the GitHub backend, leading to the refresh token functionality not being triggered.

This issue occurs because the GitHub backend uses a different key (`expires`[^1]) than the one used by GitHub (and by extension, the OAuth2 RFC[^2]), which is `expires_in`[^3]. This misconfiguration results in the expiration value always being `None`[^4], preventing the refresh functionality from being called[^5].

In order for the refresh functionality, that is already present in the `UserMixin`[^6] and `BaseOAuth2` [^7], to be utilized by the GitHub backend, we have to also include the `refresh_token` key in the `extra_data` [^1]. This ensures that we always store it and use it as outlined by GitHub[^3].

I’ll open a pull request handling the issue.

[^1]: https://github.com/python-social-auth/social-core/blob/5899cb92ba1c3799a1c614f9b444c490bee2e25c/social_core/backends/github.py#L26
[^2]: https://github.com/python-social-auth/social-core/issues/40
[^3]: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens#refreshing-a-user-access-token-with-a-refresh-token
[^4]: https://github.com/python-social-auth/social-core/blob/5899cb92ba1c3799a1c614f9b444c490bee2e25c/social_core/backends/base.py#L124-L153
[^5]: https://github.com/python-social-auth/social-core/blob/5899cb92ba1c3799a1c614f9b444c490bee2e25c/social_core/storage.py#L100-L104
[^6]: https://github.com/python-social-auth/social-core/blob/5899cb92ba1c3799a1c614f9b444c490bee2e25c/social_core/storage.py#L46-L55
[^7]: https://github.com/python-social-auth/social-core/blob/5899cb92ba1c3799a1c614f9b444c490bee2e25c/social_core/backends/oauth.py#L454-L461

gmargaritis:
Resolves #932

Ensure the correct key is used for access_token expiration in the GitHub backend's extra_data, and save the refresh_token.

Previously, the expiration of the access_token was not stored, causing the refresh_token functionality to be skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Handle case where user has not registered a `family-name`  with ORCID
+- Fix access token expiration and refresh token handling in GitHub backend
 
 ## [4.5.4](https://github.com/python-social-auth/social-core/releases/tag/4.5.4) - 2024-04-25
 
--- a/social_core/backends/github.py
+++ b/social_core/backends/github.py
@@ -23,7 +23,12 @@ class GithubOAuth2(BaseOAuth2):
     REDIRECT_STATE = False
     STATE_PARAMETER = True
     SEND_USER_AGENT = True
-    EXTRA_DATA = [("id", "id"), ("expires", "expires"), ("login", "login")]
+    EXTRA_DATA = [
+        ("id", "id"),
+        ("expires_in", "expires"),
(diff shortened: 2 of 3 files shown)


nijel:

PR #915, opened 2024-05-18
fix linkedin userdetails backend

MohamedAhmed412000:
## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers
why we should accept this pull request. If it fixes a bug or resolves a feature
request, be sure to link to that issue.

## Types of changes

Please check the type of change your PR introduces:

- [ ] Release (new release request)
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (PEP8, lint, formatting, renaming, etc)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build related changes (build process, tests runner, etc)
- [ ] Other (please describe):

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask. We're here to
help! This is simply a reminder of what we are going to look for before merging
your code._

- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added documentation to https://github.com/python-social-auth/social-docs

## Other information

Any other information that is important to this PR such as screenshots of how
the component looks before and after the change.

PR #950, opened 2025-01-01
Add passing tests for linkedin OAuth updates

offbyone:
This is a follow-up to and addition onto https://github.com/python-social-auth/social-core/pull/915

All that the preceding PR includes is part of it, plus tests passing. Props to @MohamedAhmed412000 for the initial work

It'll supersede #915 and #934

### DIFF
--- a/social_core/backends/linkedin.py
+++ b/social_core/backends/linkedin.py
@@ -66,17 +66,7 @@ class LinkedinOAuth2(BaseOAuth2):
     ]
 
     def user_details_url(self):
-        # use set() since LinkedIn fails when values are duplicated
-        fields_selectors = list(
-            set(
-                ["sub", "given_name", "family_name", "email"]
-                + self.setting("FIELD_SELECTORS", [])
-            )
-        )
-        # user sort to ease the tests URL mocking
-        fields_selectors.sort()
-        fields_selectors = ",".join(fields_selectors)
-        return self.USER_DETAILS_URL.format(projection=fields_selectors)
+        return self.USER_DETAILS_URL
 
     def user_emails_url(self):
         return self.USER_EMAILS_URL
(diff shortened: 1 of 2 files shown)


nijel:
The failing tests are affected by https://github.com/gabrielfalcao/HTTPretty/issues/484, but as they pass on 3.8 I assume this is safe to merge.

tuky:
@nijel I would like to suggest to release such changes as major version bumps. I understand that this change is needed for new apps to work. However, this change also killed OAuth with our old linkedin app. Is there any information available, how transition old apps to this new flow? What settings have to be adjusted? This is our config that now breaks all caused by a **patch** version bump which IMO is utterly wrong. M$ so far has not forced us to do something.

```python
SOCIAL_AUTH_LINKEDIN_OAUTH2_SCOPE = ['r_liteprofile', 'r_emailaddress']
SOCIAL_AUTH_LINKEDIN_OAUTH2_FIELD_SELECTORS = ['emailAddress', 'profilePicture(displayImage~:playableStreams)']

SOCIAL_AUTH_LINKEDIN_OAUTH2_EXTRA_DATA = [
    ('id', 'id'),
    ('firstName', 'first_name'),
    ('lastName', 'last_name'),
    ('emailAddress', 'email_address'),
    ('profilePicture', 'profile_picture'),
]
```

social_core/backends/linkedin.py
```diff
@@ -48,14 +48,14 @@ class LinkedinOAuth2(BaseOAuth2):
...
-    USER_DETAILS_URL = "https://api.linkedin.com/v2/me?projection=({projection})"
+    USER_DETAILS_URL = "https://api.linkedin.com/v2/userinfo?projection=({projection})"
     USER_EMAILS_URL = (
         "https://api.linkedin.com/v2/emailAddress"
         "?q=members&projection=(elements*(handle~))"
     )
     ACCESS_TOKEN_METHOD = "POST"
     REDIRECT_STATE = False
-    DEFAULT_SCOPE = ["r_liteprofile"]
+    DEFAULT_SCOPE = ["email", "profile", "openid"]
```
yuekui, line 58: These scopes are for [OpenID Connect](https://learn.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin-v2?context=linkedin%2Fconsumer%2Fcontext#authenticating-members), not OAuth2. We shouldn't make such changes in `LinkedinOAuth2` class

nijel: `r_liteprofile` was not working, see https://github.com/python-social-auth/social-core/issues/922

Issue #949, opened 2025-01-01
Add type annotations to clarify some return types


offbyone:
I would like to add some type annotations to the project to support type checking with pyright when writing extensions. The specific issue I'm running into is described in https://github.com/microsoft/pyright/issues/9648, which is manifesting when I attempt to override `authorization_url()` on a custom OAuth implementation to return a dynamic string. 

I don't see any existing issues for this, so I'm creating it here. I'm happy to do the work.

I expect it to look something like this:

1. Add validation of types via pyright to the CI process
2. Add type annotations incrementally to some base types
3. Add instructions to the contribution docs recommending more types be added.

## Related 

* #946 -- I think this is likely to help with annotations in some cases, although not all; I think we can start without dropping 3.8/3.9

nijel:
This would be definitely welcome. I have no experience with pyright, but I guess it uses the same type annotations as mypy.

offbyone:
It does, for the most part; they both have different capabilities with regards to inference, and a few quirks, but for most things they're identical. I'd be happy to include a workflow that checks with mypy in the plan.

nijel:
Great. I've done some cleanups today, so Python 3.8 support is gone, and we're linting and formatting code using ruff.

PR #986, opened 2025-01-10
Type Annotate the *heck* out of social-core

offbyone:
## Proposed changes

This is implementing the type checking I suggested in #949. 

## Types of changes

- [x] Other (please describe):

## Checklist

your code._

## Other information

This touches most of social-core, and entails some compromises due to overrides that pyright really does not like. These can be broken down into a few categories:

* cases where a value can be `None` but is being treated as not, because it's set outside of the scope of the method. In these cases, I have added `assert <var>` statements to tell the type checker that the value is non-null; this code would have raised an exception at this point anyway
* method overrides that are incompatible with the base class. `get_user_id()` and `get_user_details()` and the like are particularly egregious in this regard; I have added type:ignore comments in these cases.
* Incorrect types in library values. **These probably bugs**. Look for `TODO` comments on these; I didn't try fix them here out of an abundance of caution, since I don't use those authentication providers.
* the Model mixins all used `get()` methods, which actually led to type confusion, since sometimes more than one was mixed in. I renamed `get()` on `Nonce` to `get_nonce()` and on `Association` to `get_association()` in those cases.
* The signature types are definitely wrong somehow; `public_key = RSAAlgorithm.from_jwk(...)` is _not_ returning a public key, it's returning an allowed keys type, at least with modern pyca-cryptography; it is probably wrong, but it's working. type:ignore'd
* Annotated several return types that were mis-inferred. Often pyright would infer a literal return type, because pyright is overly picky.

### DIFF
--- a/social_core/utils.py
+++ b/social_core/utils.py
@@ -10,7 +10,7 @@ from urllib.parse import unquote, urlencode, urlparse, urlunparse
 
 import requests
 from requests.adapters import HTTPAdapter
-from requests.packages.urllib3.poolmanager import PoolManager
+from urllib3.poolmanager import PoolManager
 
 import social_core
 
@@ -34,7 +34,7 @@ class SSLHttpAdapter(HTTPAdapter):
         self.ssl_protocol = ssl_protocol
         super().__init__()
 
-    def init_poolmanager(self, connections, maxsize, block=False):
+    def init_poolmanager(self, connections, maxsize, block=False, **pool_kwargs):
         self.poolmanager = PoolManager(
             num_pools=connections,
             maxsize=maxsize,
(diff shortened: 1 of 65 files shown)


social_core/utils.py
```diff
@@ -10,7 +10,7 @@
 
 import requests
 from requests.adapters import HTTPAdapter
-from requests.packages.urllib3.poolmanager import PoolManager
+from urllib3.poolmanager import PoolManager
```
nijel, line 13: Maybe this is what breaks the testsuite? At least I don't see anything else what could be suspicious. Looking at the code I wonder if SSLHttpAdapter is still needed these days. I've tried to remove it in https://github.com/python-social-auth/social-core/pull/987.

nijel (commented):
Can you please constantly use future typing with `Type | None` syntax instead of mixing it with `Optional[Type]` in some files?

offbyone:
No... I just accidentally left in a test assertion. Fixing...

offbyone:
> Can you please constantly use future typing with `Type | None` syntax instead of mixing it with `Optional[Type]` in some files?

Happy to; that was accidental in the early steps, where I forgot that we had Python new enough to use it. I'll revise ASAP.

social_core/utils.py
offbyone: No... I just accidentally left in a test assertion. Fixing...

nijel:

PR #1013, opened 2025-02-03
fix: update Azure B2C base URL to allow custom domains

### LINKED ISSUE #809: Incorrect base url for azuread b2c backend
As per this [documentation](https://learn.microsoft.com/en-us/azure/active-directory-b2c/b2clogin#what-endpoints-does-this-apply-to), I think the base URL for backend "azuread-b2c-oauth2" for custom domain should be 

**BASE_URL** = "https://{authority_host}/{tenant_name}.onmicrosoft.com"

but not

**BASE_URL** = "https://{tenant_name}.{authority_host}/{tenant_name}.onmicrosoft.com"

From Azure documentation:

> With Azure AD B2C [custom domain](https://learn.microsoft.com/en-us/azure/active-directory-b2c/custom-domain) the corresponding updated endpoint would look like:
> 
> https://login.contoso.com/<tenant-name>.onmicrosoft.com/<policy-name>/oauth2/v2.0/authorize
> https://login.contoso.com/<tenant-name>.onmicrosoft.com/oauth2/v2.0/authorize?p=<policy-name>

Danyal-Faheem:
fixes #809.

The Authority Host can be defined to be a custom domain as well using front door in Azure. With this, the tenant name is not prefixed to the authority host in the base url.

We update this by prefixing the tenant name with the default authority host only if the authority host is not defined.

### DIFF
--- a/social_core/backends/azuread_b2c.py
+++ b/social_core/backends/azuread_b2c.py
@@ -64,7 +64,7 @@ class AzureADB2COAuth2(AzureADOAuth2):
 
     @property
     def authority_host(self):
-        return self.setting("AUTHORITY_HOST", f"{tenant_name}.b2clogin.com")
+        return self.setting("AUTHORITY_HOST", f"{self.tenant_name}.b2clogin.com")
 
     @property
     def tenant_name(self):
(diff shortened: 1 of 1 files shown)


nijel:

PR #1015, opened 2025-02-07
Google One Tap implementation

kviktor:
## Proposed changes

The changes adds Google One Tap as a backend

## Checklist

- [x] I have added documentation to https://github.com/python-social-auth/social-docs

## Other information

You can test the changes with this dummy project https://github.com/kviktor/django-onetap-example

I haven't yet added tests or documentation but if the changes look okay I'll add them too.

I have 2 questions tho I'm not exactly sure about, maybe someone can help out
- I've added google-auth as a dependency, technically it's just downloading Google's cert files then verifying the jwt token however I didn't want to reimplement the wheel for it and it's a bit safer this way imo
- as it requires a new dependency I've added a new Python module for it, is that ok? (I can move it into `google.py` with catching ImportError and raising an exception if the library is missing in the class)

Docs at https://github.com/python-social-auth/social-docs/pull/256

### DIFF
--- a/social_core/backends/google_onetap.py
+++ b/social_core/backends/google_onetap.py
@@ -1,0 +1,53 @@
+import urllib3
+from google.auth.transport.urllib3 import Request
+from google.oauth2 import id_token
+
+from social_core.backends.base import BaseAuth
+from social_core.backends.google import BaseGoogleAuth
+from social_core.exceptions import AuthException, AuthTokenError
+
+
+class GoogleOneTap(BaseGoogleAuth, BaseAuth):
+    name = "google-onetap"
+    CSRF_KEY = "g_csrf_token"
+    CREDENTIAL_KEY = "credential"
+
+    def auth_url(self):
+        raise AuthException(self, "Cannot start login flow for Google One Tap")
+
+    def verify_csrf(self, request):
+        csrf_token_body = self.data.get(self.CSRF_KEY)
+        csrf_token_cookie = request.COOKIES.get(self.CSRF_KEY)
+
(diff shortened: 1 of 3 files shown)


social_core/backends/google_onetap.py
```diff
@@ -0,0 +1,53 @@
...
+        csrf_token_body = self.data.get(self.CSRF_KEY)
+        csrf_token_cookie = request.COOKIES.get(self.CSRF_KEY)
+
+        if not csrf_token_body:
+            raise AuthTokenError(self, "Missing csrf token from response")
+
+        if not csrf_token_cookie:
+            # csrf_token_cookie can be missing due to
+            # https://issuetracker.google.com/issues/226157137
+            return
```
nijel, line 28: This makes the whole CSRF verification mostly useless as it does not verify requests without cookies.

nijel (commented):
Looks okay at first look. Can you please add tests and documentation?

social_core/backends/google_onetap.py
kviktor: I agree, I was thinking of making it configurable so users who have everything on a single domain have a sane default and users with who have the auth part on a different one can disable it, I'll add that with the tests.

kviktor:
@nijel I've added some tests, because it's not on oauth flow they are a bit different.

Also added docs at https://github.com/python-social-auth/social-docs/pull/256

Also switched from urllib3 to requests as I've just noticed this library already depends on it and that requirement doesn't need that weird packaging library workaround.

PR #1031, opened 2025-03-05
feat: allow to configure JWT leeway in open_id_connect and increase default to 1 second

jirutka:
## Proposed changes

The default value defined in the jwt module is 0 seconds, so when the time is off just by a few milliseconds, the token validation fails with error "The token is not yet valid (iat)".

## Checklist

   - This feature is already in the azuread_b2c and mediawiki backends, but there are no tests for it, so I don’t have any reference to based it on and don’t have time to figure it out myself.
    - I don’t see any documentation for the generic open_id_connect backend in https://github.com/python-social-auth/social-docs/tree/master/docs/backends.

### DIFF
--- a/social_core/backends/open_id_connect.py
+++ b/social_core/backends/open_id_connect.py
@@ -61,7 +61,7 @@ class OpenIdConnectAuth(BaseOAuth2):
     USERNAME_KEY = "preferred_username"
     JWT_ALGORITHMS = ["RS256"]
     JWT_DECODE_OPTIONS: dict[str, Any] = {}
-    JWT_LEEWAY : float = 1.0  # seconds
+    JWT_LEEWAY: float = 1.0  # seconds
     # When these options are unspecified, server will choose via openid autoconfiguration
     ID_TOKEN_ISSUER = ""
     ACCESS_TOKEN_URL = ""


jirutka:
No, `ID_TOKEN_MAX_AGE` has a different purpose. As you can see in the code base, default value for `ID_TOKEN_MAX_AGE` is 600 and 3600 seconds, that would be too high for the leeway parameter. `JWT_LEEWAY` is used in JWT validation to allow a small amount of clock drift when verifying token expiration (exp), not-before (nbf), and issued-at (iat) claims.

nijel:
> that would be too high for the leeway parameter. 

Having a wrong value doesn't mean it was not built with the same purpose. Apparently, `ID_TOKEN_MAX_AGE` is used to verify the `iat` claim (and is not applied to the `nbf` claim):

https://github.com/python-social-auth/social-core/blob/2b2a7e143e43b40a45ec131964271a4d7c931ac0/social_core/backends/open_id_connect.py#L156-L162

If this validation is now done inside the JWT library, the additional logic probably can be removed and consolidated across the backends:

* `OpenIdConnectAuth.validate_claims` seems to duplicate the 
* `LinkedinOpenIdConnect` seems to have a copy of this as well
* `MediaWiki` is passing leeway parameter to `jwt.decode` and doing manual validation later (but this one is not a `OpenIdConnect` subclass)

PR #1048, opened 2025-03-14
fix(oauth2): ACCESS_TOKEN_METHOD defaults to POST

### LINKED ISSUE #284: ACCESS_TOKEN_METHOD should default to POST instead of GET
According to the RFC (https://tools.ietf.org/html/rfc6749#section-3.2) the client MUST use POST method when making access token requests.  The default value for ACCESS_TOKEN_METHOD conflicts with this (at least with OAUTH2).

https://github.com/python-social-auth/social-core/blob/563d0e5e97e9c5e6289609d09d2e7bf7c2f3165d/social_core/backends/oauth.py#L31

nijel:
This is possibly a breaking change for a few backends but makes the code follow RFC and removes need for every backend to override this.

Fixes #284

## Checklist

your code._

## Other information

### DIFF
--- a/social_core/backends/oauth.py
+++ b/social_core/backends/oauth.py
@@ -46,7 +46,7 @@ class OAuthAuth(BaseAuth):
 
     AUTHORIZATION_URL = ""
     ACCESS_TOKEN_URL = ""
-    ACCESS_TOKEN_METHOD = "GET"
+    ACCESS_TOKEN_METHOD = "POST"
     REVOKE_TOKEN_URL = None
     REVOKE_TOKEN_METHOD = "POST"
     ID_KEY = "id"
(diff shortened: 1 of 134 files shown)


social_core/backends/oauth.py
```diff
@@ -46,7 +46,6 @@ class OAuthAuth(BaseAuth):
 
     AUTHORIZATION_URL = ""
     ACCESS_TOKEN_URL = ""
-    ACCESS_TOKEN_METHOD = "GET"
```
atodorov, line 49: The `BaseAuth` doesn't appear to have an `ACCESS_TOKEN_METHOD` attribute so unless I am missing something this may end-up being undefined in some cases and lead to errors.

Instead of deleting it, why not explicitly set the value to `POST` ? (added bonus it will be visible to anyone looking).

atodorov: Ahhh, tests failed with exactly this error 
```
AttributeError: 'GithubOAuth2' object has no attribute 'ACCESS_TOKEN_METHOD'
```

nijel: Yes, I reverted this edit after mass removing these from the other backends, but forgot to commit that part :facepalm:.

nijel:
Merged. As mentioned, this might be a breaking change for some backends not following RFC. In case you run into this, set `ACCESS_TOKEN_METHOD` to `GET` for the affected backend.

PR #1063, opened 2025-03-24
feat: use responses instead of HTTPretty

### LINKED ISSUE #1024: Replace httpretty with a maintained alternative
Seems like there is not much interest in fixing https://github.com/gabrielfalcao/HTTPretty/issues/484 and effectively blocking support for recent urllib3 (see https://github.com/python-social-auth/social-core/pull/997), so we should examine alternatives.

At Weblate we use [responses](https://pypi.org/project/responses/) which is maintained by Sentry, so it is quite likely to stay maintained, but I'm fine with other suggestions.

nijel:
## Proposed changes

This is a mantained alternative that will not block urllib3 upgrades. The downside of this is that it mocks at requests level, not at urllib3 level making it impossible to mock OpenId tests.

Fixes #1024

## Checklist

your code._

## Other information

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,11 +6,9 @@ requires = ["setuptools>=78.0.2"]
 dev = [
   "flake8>=5.0.4",
   "pytest>=4.5",
-  "httpretty~=1.1.0",
+  "responses==0.25.7",
   "coverage>=3.6",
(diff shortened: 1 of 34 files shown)


nijel:
I've type annotated newly added helper and that lead to a chain of adding type annotations around the codebase :smile:.

PR #1099, opened 2025-04-07
Add the OpenIdConnect-based backend for Fedora, and deprecate the OpenID-based one

abompard:
Fedora Infrastructure is deprecating the OpenID login method, soon it will stop working.

Ref: https://fedoramagazine.org/end-of-openid-authentication-in-fedora-account-system/

### DIFF
--- a/social_core/backends/fedora.py
+++ b/social_core/backends/fedora.py
@@ -4,9 +4,34 @@ Fedora OpenId backend, docs at:
 """
 
 from .open_id import OpenIdAuth
+from .open_id_connect import OpenIdConnectAuth
+
+
+class FedoraOpenIdConnect(OpenIdConnectAuth):
+    """
+    Fedora OpenID Connect backend.
+
+    To use it, you need to set the SOCIAL_AUTH_FEDORA_OIDC_KEY and
+    SOCIAL_AUTH_FEDORA_OIDC_SECRET configuration variables to the client_id and
+    client_secret values that the Fedora Infrastructure gave you.
+    """
+
+    name = "fedora-oidc"
+    USERNAME_KEY = "nickname"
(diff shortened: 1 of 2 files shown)


social_core/backends/fedora.py
```diff
@@ -4,9 +4,35 @@
...
+class FedoraOpenIdConnect(OpenIdConnectAuth):
+    """
+    Fedora OpenID Connect backend.
+
+    To use it, you need to set the SOCIAL_AUTH_FEDORA_KEY and
+    SOCIAL_AUTH_FEDORA_SECRET configuration variables to the client_id and
+    client_secret values that the Fedora Infrastructure gave you.
+    """
+
+    name = "fedora"
```
nijel, line 19: The backend name should be different, for example:

```suggestion
    name = "fedora-oidc"
```

nijel: Or is this really expected to be a drop-in replacement including same user identifiers?

nijel: What I meant is user associations. For example, with OpenID I have:

image

Will be the UID provided by OIDC same as with Open ID? If not, then the backend should get a different name so that it doesn't try to match OpenID UIDs against OIDC ones.

abompard: Ah I see, no UIDs will be taken from the `sub` attribute according to the OpenID Connect protocol. I'll change the name, thanks for pointing this out.

nijel (approved):
Looks good now, thanks. Can you please also update https://github.com/python-social-auth/social-docs/blob/master/docs/backends/fedora.rst? Preferably including info on getting credentials.

abompard:
Done! I've also updated the docstring as the setting names change with the backend name.

nijel:

PR #1103, opened 2025-04-09
feat: Add JWT leeway setting to KeycloakOAuth2 backend

### LINKED ISSUE #1102: ImmatureSignatureError The token is not yet valid
Hello,

The problem I'm facing is identical to the one solved in #685 but with a different backend, Keycloak.

The solution that is proposed, to add `SOCIAL_AUTH_JWT_LEEWAY` in the Django settings does not work as leeway is not implemented in the backend code:

```python
def user_data(self, access_token, *args, **kwargs):  # pylint: disable=unused-argument
    """Decode user data from the access_token

    You can specialize this method to e.g. get information
    from the Keycloak backend if you do not want to include
    the user information in the access_token.
    """

    return jwt.decode(
        access_token,
        key=self.public_key(),
        algorithms=self.algorithm(),
        audience=self.audience()
    )
```

I checked the AzureAD B2C backend and the jwt.decode is implementing the leeway in the following way:

```python
return jwt_decode(
    id_token,
    key=key,
    algorithms=["RS256"],
    audience=self.setting("KEY"),
    leeway=self.setting("JWT_LEEWAY", default=0),
)
```

I will submit a PR to port this to the Keycloak backend.

mageo:
Fixes #1102 by adding support for a `JWT_LEEWAY` setting in the Keycloak backend.

### DIFF
--- a/social_core/backends/keycloak.py
+++ b/social_core/backends/keycloak.py
@@ -133,6 +133,7 @@ class KeycloakOAuth2(BaseOAuth2):  # pylint: disable=abstract-method
             key=self.public_key(),
             algorithms=self.algorithm(),
             audience=self.audience(),
+            leeway=self.setting("JWT_LEEWAY", default=0),
         )
 
     def get_user_details(self, response):


nijel:
I'd really prefer to have a generic solution here instead of ad-hoc solutions per backend. See also https://github.com/python-social-auth/social-core/pull/1031

Issue #1124, opened 2025-04-24
Provide support to optional parameter in OpenID Connect backend


rgaiacs:
The ["OpenID Connect Core 1.0" specification](https://openid.net/specs/openid-connect-core-1_0.html) defines a few optional parameters to be used during the [authentication request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest). This optional parameters are

- [x] `nonce`
- [ ] `display`
- [ ] `prompt`
- [ ] `max_age`
- [ ] `ui_locales`
- [ ] `id_token_hint`
- [ ] `login_hint`
- [ ] `acr_values`

The implemented [Open Id Connect backend](https://github.com/python-social-auth/social-core/blob/7066da93129eba8a903153bcba6e6742e57fc640/social_core/backends/open_id_connect.py) already implements `nonce` in https://github.com/python-social-auth/social-core/blob/7066da93129eba8a903153bcba6e6742e57fc640/social_core/backends/open_id_connect.py#L133

It will be great to provide users the option to configure other optional parameters by [defining some settings](https://python-social-auth.readthedocs.io/en/latest/configuration/settings.html#settings-name), for example for prompt we could have

```
SOCIAL_AUTH_OIDC_PROMPT = 'select_account'
```

nijel:
Partly implemented in https://github.com/python-social-auth/social-core/pull/1127

PR #1127, opened 2025-04-24
Add support to SOCIAL_AUTH_OIDC_PROMPT and others

rgaiacs:
This is related to https://github.com/python-social-auth/social-core/issues/1124.

This is my first PR to this project. Feedback is welcome, specially regarding

- [ ] how to trigger a logging warning?
- [ ] how to add a test to this new feature?

If this PR looks good, I will submit others to complete https://github.com/python-social-auth/social-core/issues/1124.


rgaiacs:
A PR with documentation was sent, see https://github.com/python-social-auth/social-docs/pull/288.

social_core/backends/open_id_connect.py
```diff
@@ -131,6 +133,16 @@ def auth_params(self, state=None):
         """Return extra arguments needed on auth process."""
         params = super().auth_params(state)
         params["nonce"] = self.get_and_store_nonce(self.authorization_url(), state)
+
+        prompt = self.setting("PROMPT", default=self.PROMPT)
+        is_prompt_valid = True
+        for prompt_token in prompt.split():
+            if prompt_token not in ("none", "login", "consent", "select_account"):
+                is_prompt_valid = False
```
nijel, line 141: If the configuration is invalid, it should IMHO fail with an exception.

social_core/backends/open_id_connect.py
```diff
@@ -131,6 +133,16 @@ def auth_params(self, state=None):
...
+                break
+        if is_prompt_valid:
```
nijel, line 143: Should empty prompt be included as well?

nijel (commented):
If warnings were meant to indicate errors in the configuration, I think it's better to fail directly. Nobody will notice warnings.

As for the tests, it could check whether the configured parameters are included in the request to the authentication server.

rgaiacs:
Thanks for the feedback. I expanded to match the list of parameters explicitly mentioned in the [Authentication Request section](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).

I don't implemented all parameters because of my limited knowledge of OpenID.

social_core/backends/open_id_connect.py
```diff
@@ -131,6 +139,53 @@ def auth_params(self, state=None):
         """Return extra arguments needed on auth process."""
         params = super().auth_params(state)
         params["nonce"] = self.get_and_store_nonce(self.authorization_url(), state)
+
+        display = self.setting("DISPLAY", default=self.DISPLAY)
+        if display is not None:
+            if not display:
+                raise ValueError("OpenID Connect display value cannot be empty string.")
```
nijel, line 146: Maybe it would better to introduce a new exception in https://github.com/python-social-auth/social-core/blob/master/social_core/exceptions.py and use it instead (`AuthMissingParameter` is closest, but IMHO not really a good fit here).

nijel:

PR #1147, opened 2025-05-06
Revert LinkedIn OAuth2

yuekui:
The previous PR #950 mistakenly replaced the OAuth2 endpoint with the OpenID Connect one. Users who use OpenID Connect should use `LinkedinOpenIdConnect`. For existing LinkedIn OAuth2 users like myself, `LinkedinOAuth2` should still work. 

Just to clarify: LinkedIn has **NOT** removed the `r_liteprofile` scope for OAuth2 users — it's still functioning properly. If anyone is having issues with scopes, please check the “Products” tab in your LinkedIn App Setup page.

For example, here we have both OAuth2 and OpenID Connect enabled:
image
image

### DIFF
--- a/social_core/backends/linkedin.py
+++ b/social_core/backends/linkedin.py
@@ -67,7 +67,7 @@ class LinkedinOAuth2(BaseOAuth2):
     def user_details_url(self):
         # use set() since LinkedIn fails when values are duplicated
         fields_selectors = list(
-            set(["id", "firstName", "lastName"] + self.setting("FIELD_SELECTORS", []))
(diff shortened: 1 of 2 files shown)


bogdandm:
Hello, is it possible to create release with this fix? LinkedinOAuth2 does not work for our service and we have to copy-paste whole LinkedinOAuth2 from previous version as it critical to our users to be able to login into their accounts

nijel:
I wanted to merge https://github.com/python-social-auth/social-core/pull/1183 before the release, but it is not yet ready...

Issue #684, opened 2022-05-08
AzureAD: use objectId instead of UPN as unique identifier


Exordian:
### Expected behaviour

Use Azure objectId as identifier.

### Actual behaviour

UPN is used as user identifier.

### What are the steps to reproduce this issue?

Use Azure AD authentication.

### Any logs, error output, etc?

-
### Any other comments?

The UPN should reflect the primary mail as recommended by Microsoft [1]. The e-mail address might change over time due to a new name of the person, merger and acquisitions or similar actions.

Microsoft recommends to use the objectId as unique identifier.

[1] https://docs.microsoft.com/en-us/azure/active-directory/hybrid/howto-troubleshoot-upn-changes

svdHero:
@Exordian That would be `oid` then and not `objectId`, wouldn't it?

joachimBurket:
@svdHero  yes it would be the `oid`. 
I also saw that the `upn`, `preferred_username`, or `sub` claims are used as the unique identifier (depending on the class), and I also thought that the `oid` was the claim recommended by Microsoft.

joachimBurket:
I can make a PR to change it, but as I'm just starting to use this library, I'm not sure what it would imply on existing applications.

nijel:
Just changing `upn` to `oid` would break existing associations. I don't think there is currently an existing way to migrate to a different user identifier gracefully.

PR #1183, opened 2025-06-09
Feature/azure oid backend

stealthycole:
This is just a slightly modified version of the azure_tenant backend that uses OID instead of sub claim as the UID. Allows admins to more-safely switch to a version of Azure SSO that uses OID.

Issue: https://github.com/python-social-auth/social-core/issues/684

### DIFF
--- a/social_core/backends/azuread_oid.py
+++ b/social_core/backends/azuread_oid.py
@@ -42,6 +42,7 @@ class AzureADOIDOAuth2(AzureADTenantOAuth2):
         """Use account oid as unique id."""
         return response.get("oid")
 
+
 class AzureADV2OIDOAuth2(AzureADOIDOAuth2):
     name = "azuread-v2-OID-oauth2"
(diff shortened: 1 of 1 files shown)


social_core/backends/azuread_oid
```diff
@@ -0,0 +1,137 @@
...
+            certificate = self.get_certificate(key_id)
+
+            return jwt_decode(
+                id_token,
+                key=certificate.public_key(),  # type: ignore[reportArgumentType]
+                algorithms=["RS256"],
+                audience=self.setting("KEY"),
+            )
+        except (DecodeError, ExpiredSignatureError) as error:
+            raise AuthTokenError(self, error)
```
nijel: Why is this duplicating so many logic? I would expect that it would only differ in `get_user_id`.

stealthycole: Good point, I've trimmed that down.

nijel:
Can you please add tests and documentation to https://github.com/python-social-auth/social-docs?

PR #1212, opened 2025-07-29
fix(backends): use HTTP headers to authenticate for Gitea

### LINKED ISSUE #1211: The Gitea OAuth2 adapter will break with Forgejo instance
In 2023, Gitea and Forgejo marked several authentication methods as deprecated, which are used by the social-core oauth2 provider.

* Commit in Forgejo: https://codeberg.org/forgejo/forgejo/commit/4e879fed90665331d2a57e5abee9e0f02372c470
* Commit in Gitea: https://github.com/go-gitea/gitea/commit/4e879fed90665331d2a57e5abee9e0f02372c470

With Forgejo v12.0.0, the deprecated methods have initially been removed. However, we realized that some tools, including social-core, rely on the deprecated authentication methods. Forgejo v12.0.1 has re-enabled the behaviour. It is currently subject to discussion on when the methods will be retired forever, but it might happen with Forgejo v13.

The reasoning for removal is to ensure that access tokens are not part of the URL and are not logged or otherwise exposed inadvertently.

It would be great if social-core would update the auth methods to something that will be supported in Gitea and Forgejo going forward, there are multiple options available.

(BTW, it would also be useful to allow people to define a dedicated Forgejo provider, even if it's technically a reference to the Gitea one. It would be useful in case the code needs to diverge in the future; users will have already chosen the appropriate one for their use case.)

nijel:
The query string authentication is deprecated and will be removed.

Fixes #1211

### DIFF
--- a/social_core/backends/gitea.py
+++ b/social_core/backends/gitea.py
@@ -46,5 +46,6 @@ class GiteaOAuth2(BaseOAuth2):
     def user_data(self, access_token, *args, **kwargs):
         """Loads user data from service"""
         return self.get_json(
-            self.api_url("/api/v1/user"), params={"access_token": access_token}
+            self.api_url("/api/v1/user"),
+            headers={"Authorization": f"token {access_token}"},
         )
--- a/social_core/backends/gitea.py
+++ b/social_core/backends/gitea.py
@@ -47,5 +47,5 @@ class GiteaOAuth2(BaseOAuth2):
         """Loads user data from service"""
         return self.get_json(
             self.api_url("/api/v1/user"),
-            headers={"Authorization": f"token {access_token}"},
+            headers={"Authorization": f"Bearer {access_token}"},
         )


social_core/backends/gitea.py
```diff
@@ -46,5 +46,6 @@ def get_user_details(self, response):
     def user_data(self, access_token, *args, **kwargs):
         """Loads user data from service"""
         return self.get_json(
-            self.api_url("/api/v1/user"), params={"access_token": access_token}
+            self.api_url("/api/v1/user"),
+            headers={"Authorization": f"token {access_token}"},
```
viceice:
```suggestion
            headers={"Authorization": f"Bearer {access_token}"},
```

`Bearer` is more standard conform and supported / preferred by Gitea / Forgejo.

nijel: The Swagger specification doesn't mention `Bearer` at all, it only mentions:

> API tokens must be prepended with "token" followed by a space.

See https://codeberg.org/forgejo/forgejo/src/commit/d3ca92df79555007f102f930e721a68d96eb29cf/templates/swagger/v1_json.tmpl#L30108

viceice: we should fix that. it supports both: 

https://codeberg.org/forgejo/forgejo/src/commit/e3bfa5133f40c5d4819cee402b772e0bb94affa9/services/auth/oauth2.go#L141

https://github.com/go-gitea/gitea/blob/e02b47d2f32736d8c032e9daa602a143ee8a2548/modules/auth/httpauth/httpauth.go#L42

PR #1240, opened 2025-08-27
Auth0 OIDC backend

marius-mather:
I have been using the generic OIDC backend for Auth0 and it more or less works, so I don't think too many changes are required from the generic OIDC backend. However, I'm running into issues with the provider name etc. that that I think would be solved by having an auth0 specific implementation.

Please let me know what else would be required in terms of tests etc., I need to do a bit of further testing of the refresh tokens.


nijel:
Thanks, can you please also add it to the documentation? https://github.com/python-social-auth/social-docs

PR #1241, opened 2025-08-27
Set SEND_USER_AGENT as True on MediaWiki backend

albertoleoncio:
Sending the user-agent is now mandatory when using the Wikimedia APIs (Wikipedia, Wikidata, ...). Without it, the server returns a 403 with the following message:
`Please set a user-agent and respect our robot policy https://w.wiki/4wJS. See also T400119` ([link](https://phabricator.wikimedia.org/T400119))


nijel:
I actually don't see a reason why this should default to false, so changing the default value might be a better approach

lucaswerkmeister:
> Indeed the risk is possibly there, but I think it is more reasonable to send user-agent header by default.

Done in #1258.

PR #1255, opened 2025-09-09
Change domains for VK backend

igorgaming:
VK has started sending out such a message to everyone who has OAuth integration:

image

Eng translation:
> VKontakte switches to a domain vk.ru - now all API integrations and authorizations will be available only through it.
> 
> To ensure that your services work correctly, change their domains by September 30th, for example:
> - vk.ru/dev instead of vk.com/dev
> - oauth.vk.ru instead of oauth.vk.com
> - api.vk.ru instead of api.vk.com
> 
> If you have any questions, please contact Support: dev.vk.com/ru/support

### DIFF
--- a/social_core/backends/vk.py
+++ b/social_core/backends/vk.py
@@ -93,8 +93,8 @@ class VKOAuth2(BaseOAuth2):
 
     name = "vk-oauth2"
     ID_KEY = "id"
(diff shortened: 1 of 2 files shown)


igorgaming:
> Probably the tests need to be adjusted as well

Yeah, I'll do it a little later

PR #1259, opened 2025-09-10
OIDC: use info in id_token if not in response

lldelisle:
Dear developpers,
When using the OpenIdConnectAuth class with a ADFS IdP I got an issue because the email and `USERNAME_KEY` was not part of the response but part of the id_token.
The changes I made propose to use the info from the id_token if the key is not present in the response.
I don't know if this makes sense or if this should be a new Class.

I don't know how to make a test for this.

### DIFF
--- a/social_core/backends/open_id_connect.py
+++ b/social_core/backends/open_id_connect.py
@@ -337,6 +337,10 @@ class OpenIdConnectAuth(BaseOAuth2):
 
     def get_user_details(self, response):
         username_key = self.setting("USERNAME_KEY", self.USERNAME_KEY)
+        # populate response with id_token if needed
+        for key in [username_key, "email", "name", "given_name", "family_name"]:
+            if response.get(key) is None and self.id_token is not None:
+                response[key] = self.id_token.get(key)
         return {
             "username": response.get(username_key),
             "email": response.get("email"),
--- a/social_core/backends/open_id_connect.py
+++ b/social_core/backends/open_id_connect.py
@@ -337,7 +337,6 @@ class OpenIdConnectAuth(BaseOAuth2):
 
     def get_user_details(self, response):
         username_key = self.setting("USERNAME_KEY", self.USERNAME_KEY)
-        # populate response with id_token if needed
         user_details = {}
         for key in [username_key, "email", "name", "given_name", "family_name"]:
             if key in response:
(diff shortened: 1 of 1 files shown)


social_core/backends/open_id_connect.py
```diff
@@ -337,6 +337,10 @@ def user_data(self, access_token, *args, **kwargs):
 
     def get_user_details(self, response):
         username_key = self.setting("USERNAME_KEY", self.USERNAME_KEY)
+        # populate response with id_token if needed
+        for key in [username_key, "email", "name", "given_name", "family_name"]:
+            if response.get(key) is None and self.id_token is not None:
+                response[key] = self.id_token.get(key)
```
nijel: I'm not sure if changing the response here will not have undesired side effects. Maybe it would be better to directly build the result dict here.

lldelisle: Good catch, I changed it.

social_core/backends/open_id_connect.py
```diff
@@ -337,12 +337,19 @@ def user_data(self, access_token, *args, **kwargs):
 
     def get_user_details(self, response):
         username_key = self.setting("USERNAME_KEY", self.USERNAME_KEY)
+        # populate response with id_token if needed
```
nijel:
```suggestion
```

social_core/backends/open_id_connect.py
```diff
@@ -337,12 +337,19 @@ def user_data(self, access_token, *args, **kwargs):
...
-            "email": response.get("email"),
-            "fullname": response.get("name"),
-            "first_name": response.get("given_name"),
-            "last_name": response.get("family_name"),
+            "username": user_details.get(username_key),
+            "email": user_details.get("email"),
+            "fullname": user_details.get("name"),
+            "first_name": user_details.get("given_name"),
+            "last_name": user_details.get("family_name"),
         }
```
nijel: Wouldn't just `return user_details` work the same? It has all the keys now.

lldelisle: Not exactly, for example the key of user_details is `username_key` instead of `"username"`. Also if one key is not present neither in the response nor in the id_token, it will not be present in `user_details`.

nijel: Ah, I didn't notice this originally. Anyway not using intermediate dict would make it more readable. How about:

```py
    username_key = self.setting("USERNAME_KEY", self.USERNAME_KEY)
    def get_value(key):
        if key in response:
            return response.get(key)
        elif self.id_token is not None:
            return self.id_token.get(key)
        return None

    return {
        "username": get_value(username_key),
        "email": get_value("email"),
        "fullname": get_value("name"),
        "first_name": get_value("given_name"),
        "last_name": get_value("family_name"),
    }
```

lldelisle: Very nice suggestion. I introduced it.

Issue #1266, opened 2025-09-15
Re-implement the python-jose jwt option - verify_at_hash for when provider has not correctly configured the `at_hash` calculation


trauty-is-me:
With the change from python-jose to pyjwt the following chunk of code was not used:
```JWT_DECODE_OPTIONS = {"verify_at_hash": False}```

There was work completed to skip checking the `at_hash` value if it was not provided, however I work with a provider that has implemented it incorrectly.

The PR for this issue will reimplement this functionality and ensure that it is tested.

PR #1267, opened 2025-09-15
Reimplement option to bypass `at_hash` that was lost when changing from python-jose to pyjwt (Issue #1266)

trauty-is-me:
This adds a static field to the backend to allow bypassing the `at_hash` check even if an `at_hash` value is provided.


social_core/tests/backends/open_id_connect.py
```diff
@@ -121,12 +122,12 @@ def get_id_token(
     def prepare_access_token_body(
         self,
         client_key=None,
-        tamper_message=False,
```
nijel, line 124: Why was the argument removed when the code using it is still there? I believe it is better to list all arguments explicitly than parse them from kwargs.

trauty-is-me: I was getting pre-commit error for too many params for the function, I am happy to revert the change and increase the count or add a NOQA which ever you would prefer.

nijel: Having NOQA is probably the best approach here.

trauty-is-me: Done as NOQA.

social_core/tests/backends/open_id_connect.py
```diff
@@ -53,6 +53,9 @@ class OpenIdConnectTest(
     issuer: str  # id_token issuer
     openid_config_body: str
     key: dict[str, str]
+    allow_invalid_at_hash = False
+    skip_invalid_at_hash = False
+    access_token_kwargs: dict[str, Any]
```
trauty-is-me, line 58: typing here was required to ensure that `pyright` stops throwing errors on `prepare_access_token_body` for `tamper_message` not being a string.

social_core/tests/backends/open_id_connect.py
```diff
@@ -228,3 +234,12 @@ def test_invalid_kid(self) -> None:
         self.authtoken_raised(
             "Token error: Signature verification failed", kid="doesnotexist"
         )
+
+    def test_invalid_at_hash(self) -> None:
+        if self.skip_invalid_at_hash:
+            self.skipTest("the call doesn't match any registered mock.")
```
trauty-is-me, line 240: This is in because `einfracz`, `elixr` and `fence` do not have responses in place for a successful `do_login`

PR #1312, opened 2025-09-29
feat(saml): do not require RelayState

nijel:
The RelayState cannot be present on IdP initiated login flow and it is actually not needed in case only one IdP is configured.

### DIFF
--- a/social_core/backends/saml.py
+++ b/social_core/backends/saml.py
@@ -284,7 +284,7 @@ class SAMLAuth(BaseAuth):
                 raise AuthMissingParameter(self, "RelayState.idp")
             # Use the only configured IDP
             idp_name = next(iter(enabled_idps))
-        idp_config = self.setting("ENABLED_IDPS")[idp_name]
+        idp_config = enabled_idps[idp_name]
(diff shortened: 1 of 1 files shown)


social_core/backends/saml.py
```diff
@@ -275,8 +275,15 @@ class SAMLAuth(BaseAuth):
...
+    def get_idp(self, idp_name: str | None) -> SAMLIdentityProvider:
         """Given the name of an IdP, get a SAMLIdentityProvider instance"""
+        enabled_idps: dict[str, dict] = self.setting("ENABLED_IDPS")
+        if idp_name is None:
+            # RelayState was missing, perhaps an IdP initiated flow
+            if len(enabled_idps) != 1:
+                raise AuthMissingParameter(self, "RelayState.idp")
+            # Use the only configured IDP
+            idp_name = next(iter(enabled_idps))
         idp_config = self.setting("ENABLED_IDPS")[idp_name]
```
nijel:
```suggestion
        idp_config = enabled_idps[idp_name]
```

PR #1422, opened 2025-11-20
chore: improve type annotations

nijel:
Add type annotations and add safeguards against some unexpected values.

### DIFF
--- a/social_core/strategy.py
+++ b/social_core/strategy.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import secrets
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any, Protocol, cast
 
 from .backends.utils import get_backend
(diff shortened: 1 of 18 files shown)


social_core/strategy.py
```diff
@@ -12,7 +12,11 @@
 
 if TYPE_CHECKING:
     from .backends.base import BaseAuth
-    from .storage import BaseStorage, CodeMixin
+    from .storage import BaseStorage, CodeMixin, PartialMixin, UserProtocol
+
+
+class HttpResponseProtocol(Protocol):
+    url: str
```
andersk, line 19: Django’s `HttpResponseRedirect` does not satisfy this protocol, because its `url` property is [read-only](https://github.com/typeddjango/django-stubs/blob/5.2.8/django-stubs/http/response.pyi#L138-L139).

```suggestion
    @property
    def url(self) -> str: ...
```

The general `HttpResponse` doesn’t even have a `url` property.

- #1511

PR #1477, opened 2025-12-15
doc: add CONTRIBUTING.md

jrutila:
Added CONTRIBUTING.md doc.

It was generated by AI but I tested it by following it. Seems to work although the linting tools are generating error. Is one of them the one that should be used?

The contents from https://github.com/python-social-auth/.github/blob/bdb55771271cb4837048cd53b799436243589b45/CONTRIBUTING.md could be included or linked in this doc. That doc is just missing all the technical stuff to get the project actually running on local dev environment.


nijel:
There is already one shared for all organization repos, if it is missing something, let's update it, not create another version, which will make it impossible to maintain across repositories.

PR #1479, opened 2025-12-15
feat(fb): api version bump to 24.0

jrutila:
Facebook API v18 will be deprecated on Jan 26, 2026. 

The tests don't really run against the actual FB api? But I tested this on my own project and the logging in still worked fine with v24.


nijel:
Yes, the tests are mocked. Only a very basic API is used here and I don't think there was ever a API change that affected calls used in python-social-auth.

PR #1662, opened 2026-04-13
Catch ValueError in `sanitize_redirect`

rafsaf:
`ValueError netloc 'some-url-here' contains invalid characters under NFKC normalization` exception can be obtained if using characters like in added test.

Originally during `/login/google-oauth2?next=https://evil.garbages.com` redirect in django app using social-auth-app-django.


nijel:

PR #1682, opened 2026-04-23
update PKCE default code settings to match RFC

### LINKED ISSUE #1681: PKCE default code challenge settings don't match RFC
https://datatracker.ietf.org/doc/html/rfc7636 says the code verifier length has to be between 43-128 characters, and that the code challenge method has to be "S256" not "s256".

The current defaults in `BaseOAuth2PKCE` have code verifier length is 32 characters, and code challenge method is "s256".   When I was setting up a backend to integrate with Keycloak, I encountered issues with both of these default settings and had to change them in my custom backend, but this should probably be fixed in the base class to match what is defined in the RFC.

ragnarok56:
Updates the `BaseOAuth2PKCE` class default code challenge settings to match the RFC.  This should reduce gotchas for trying to integrate with oauth providers that are validating the code verifier length and casing on the code challenge method.

### DIFF
--- a/social_core/backends/oauth.py
+++ b/social_core/backends/oauth.py
@@ -558,8 +558,8 @@ class BaseOAuth2PKCE(BaseOAuth2):
         https://datatracker.ietf.org/doc/html/rfc7636
     """
 
-    PKCE_DEFAULT_CODE_CHALLENGE_METHOD = "s256"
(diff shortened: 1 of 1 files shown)


nijel:
Scheduled for merge, thanks for your contribution!

PR #1743, opened 2026-05-15
add VK ID backend

MeRuslan:
## Summary

Adds a new `VKIDOAuth2` backend for VK ID under the backend name `vk-id`.

VK has moved newer integrations to VK ID, which differs from the existing legacy `vk-oauth2` backend:

- uses `id.vk.ru` authorization/token/user-info endpoints
- requires OAuth 2.1 style PKCE with `code_challenge` / `code_verifier`
- returns and requires `device_id` during code exchange
- may return callback data either as flat query params or as a JSON `payload`

The existing `VKOAuth2` backend and `vk-oauth2` provider name are left unchanged for backward compatibility.

## Docs

- VK ID Web SDK: https://github.com/VKCOM/vkid-web-sdk
- VK ID no-SDK web auth: https://id.vk.ru/about/business/go/docs/ru/vkid/latest/vk-id/connection/start-integration/auth-without-sdk/auth-without-sdk-web

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- VK ID OAuth2 backend.
+
 ### Removed
 
 - Discontinued OAuth backends: AppsFuel, Beats Music, ChangeTip, Clef,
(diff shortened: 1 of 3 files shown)


nijel:
Can you please also add documentation to https://github.com/python-social-auth/social-docs?

nijel (changes requested):
Please fix linting issues and merge with current `master`.

PR #1793, opened 2026-06-16
Fix Auth0 OIDC config paths

marius-mather:
We encountered an issue where the `Auth0OpenIdConnectAuth()` backend failed to look up OIDC config correctly because of a double slash: `https://example.auth0.com//.well-known/openid-configuration`. Looks like this is because it was slightly inconsistent with the base `oidc_endpoint()` implementation, and was including a trailing slash there. I have added some code to make sure we normalize the different URLs involved better, and added tests.

P.S. I contributed this Auth0 OIDC backend, and am actively using it in our projects, so feel free to ping me in future if there's issues with it.


nijel:

PR #1829, opened 2026-06-26
fix(line): validate oauth state on callback

nijel:
Validate the returned LINE OAuth state before exchanging authorization codes to prevent login CSRF. Add regression coverage for LINE state handling and redirect URI consistency.


thientd (approved):
enforce CSRF state is good

PR #1893, opened 2026-08-04
feat(utils): add ALLOWED_REDIRECT_SCHEMES for native app redirects

### LINKED ISSUE #1892: No supported way to redirect to a private-use URI scheme (native app deep link) after login
## Summary

There is no supported way to redirect to a private-use URI scheme (a native-app deep link such as `com.example.app://oauth2redirect`) at the end of the auth flow. Setting `SOCIAL_AUTH_LOGIN_REDIRECT_URL` to one worked up to and including 4.5.4 and raises from 4.5.5 onwards.

[RFC 8252 §7.1](https://datatracker.ietf.org/doc/html/rfc8252#section-7.1) makes private-use URI scheme redirection the standard way for a native app to receive an OAuth response, so this looks like a real gap rather than just a regression.

## Root cause

Two independent places reject non-http(s) URLs.

**1. `is_url()` / `setting_url()`** — `is_url()` accepts only `http://`, `https://` and `/`, so a private-use scheme makes `setting_url()` return `None` in `do_complete()`, which then raises:

```python
url = setting_url(backend, redirect_value, "LOGIN_REDIRECT_URL")   # -> None
...
if not url:
    raise ValueError("By this point URL has to have been set")
```

**2. `sanitize_redirect()`** — since 4.8.7 it rejects the scheme outright, so even with (1) addressed the URL is dropped:

```python
if parsed_url.scheme and parsed_url.scheme not in {"http", "https"}:
    return None
```

## History

| version | behaviour |
| --- | --- |
| 4.5.4 | redirects to the deep link |
| 4.5.6 | `AssertionError: By this point URL has to have been set` |
| 5.0.2 | `ValueError: By this point URL has to have been set` |

Before 4.5.5 this worked because `setting_url()` returned `None` and the code fell through to `sanitize_redirect(...) or backend.setting("LOGIN_REDIRECT_URL")` — the `or` branch returned the setting unsanitized. That was arguably a bug, and I'm not asking for it back.

The guard arrived in 4.5.5 via #986 (commit 3cce6f3), a typing PR that doesn't mention redirects. Its stated rationale was:

> I have added `assert <var>` statements to tell the type checker that the value is non-null; this code would have raised an exception at this point anyway.

In this path it wouldn't have — it returned a redirect, so the behaviour change looks unintended.

The scheme check came separately in 4.8.7 (commit e2d5409, "Tightened redirect URL validation").

## Proposal

Add an `ALLOWED_REDIRECT_SCHEMES` setting mirroring the existing `ALLOWED_REDIRECT_HOSTS`, defaulting to `{"http", "https"}` so current behaviour is unchanged. Both `is_url()`/`setting_url()` and `sanitize_redirect()` would consult it.

That keeps the 4.8.7 hardening intact — the scheme still has to be explicitly opted into by the deployment — while giving native-app flows a supported path. Happy to open a PR if the shape looks right.

charity-detalytics:
Closes #1892.

## Problem

There is no supported way to redirect to a private-use URI scheme ([RFC 8252 §7.1](https://datatracker.ietf.org/doc/html/rfc8252#section-7.1)) at the end of the auth flow, which is how a native app receives an OAuth response. Two places reject them: `is_url()` (so `setting_url()` returns `None` and `do_complete()` raises) and `sanitize_redirect()` (which restricts schemes to http/https since 4.8.7).

## Approach

Adds an `ALLOWED_REDIRECT_SCHEMES` setting mirroring the existing `ALLOWED_REDIRECT_HOSTS`, defaulting to `("http", "https")`.

- `sanitize_redirect()` and `is_url()` take an optional `allowed_schemes` argument; omitted, behaviour is unchanged.
- `setting_url()` reads the setting from the backend and passes it down.
- The three `sanitize_redirect()` call sites in `actions.py` pass it through `get_allowed_redirect_schemes()`.

When a URL uses an explicitly allowlisted non-web scheme, the host allowlist is skipped — a private-use scheme is registered by a single native application, so the scheme itself is the trust boundary and `netloc` carries no meaning there. Web URLs are unaffected and still checked against `ALLOWED_REDIRECT_HOSTS`.

## Security

A deployment has to name a scheme before it is accepted, so the 4.8.7 hardening stays in force. Verified against `do_complete()`:

| `ALLOWED_REDIRECT_SCHEMES` | result for `com.example.app://oauth2redirect` |
| --- | --- |
| unset | `ValueError: By this point URL has to have been set` (unchanged) |
| `("http", "https", "com.example.app")` | redirects to the deep link |
| `("http", "https", "com.other.app")` | `ValueError` — the allowlist is not a blanket bypass |

The existing `test_invalid_non_web_scheme_redirect` passes untouched, which pins the default.

## Tests

Adds `SanitizeRedirectAllowedSchemesTest` and `IsUrlTest`: opted-in scheme accepted, non-allowlisted scheme rejected, `javascript:`/`ftp:` still rejected, host checks still applied to web URLs, evil URLs (`///evil.com`, backslash, control characters) still rejected, and restricting the setting can block web URLs too.

Full suite: 1134 passed locally (shopify/saml/google-onetap skipped — optional deps not installed).

### DIFF
--- a/social_core/actions.py
+++ b/social_core/actions.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Any, cast
 from urllib.parse import quote
 
 from .utils import (
+    get_allowed_redirect_schemes,
     partial_pipeline_result,
     sanitize_redirect,
     setting_url,
@@ -42,9 +43,9 @@ def _sanitize_redirect_url(backend: BaseAuth, url: str) -> str:
             *cast("list[str]", backend.setting("ALLOWED_REDIRECT_HOSTS", [])),
             backend.strategy.request_host(),
         ]
-        sanitized_url = sanitize_redirect(allowed_hosts, url) or backend.setting(
-            "LOGIN_REDIRECT_URL"
-        )
+        sanitized_url = sanitize_redirect(
+            allowed_hosts, url, get_allowed_redirect_schemes(backend)
+        ) or backend.setting("LOGIN_REDIRECT_URL")
         if sanitized_url is None:
             raise ValueError("Disallowed URL")
         url = cast("str", sanitized_url)
@@ -97,7 +98,9 @@ def do_auth(backend: BaseAuth, redirect_name: str = "next") -> HttpResponseProto
                 *cast("list[str]", backend.setting("ALLOWED_REDIRECT_HOSTS", [])),
                 backend.strategy.request_host(),
             ]
-            redirect_uri = sanitize_redirect(allowed_hosts, redirect_uri)
+            redirect_uri = sanitize_redirect(
+                allowed_hosts, redirect_uri, get_allowed_redirect_schemes(backend)
+            )
         backend.strategy.session_set(
             redirect_name, redirect_uri or backend.setting("LOGIN_REDIRECT_URL")
         )
@@ -241,7 +244,9 @@ def do_disconnect(
                 backend.strategy.request_host(),
             ]
             url = (
-                sanitize_redirect(allowed_hosts, url)
+                sanitize_redirect(
+                    allowed_hosts, url, get_allowed_redirect_schemes(backend)
+                )
                 or backend.setting("DISCONNECT_REDIRECT_URL")
                 or backend.setting("LOGIN_REDIRECT_URL")
             )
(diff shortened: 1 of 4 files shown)


nijel:
@codex review

nijel:
Scheduled for merge, thanks for your contribution!

PR #1898, opened 2026-08-06
fix(oidc): validate ID tokens on refresh

### LINKED ISSUE #589: OpenIdConnect refreshes should probably validate the ID Token
### Expected behaviour

`OpenIdConnectAuth` should reprocess the ID Token if `refresh_token` is called.

### Actual behaviour

The ID Token is nly validated whenever `OpenIdConnect.request_access_token` is called. It is not validated in the `refresh_token` case.

While the ID Token is not mandatory for refresh response, it may be provided (see https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse). The OIDC specification unfortunately does not make it clear what clients should do with the ID Token. In my opinion it is better in that case to validate it, since it is going to be used to update `user_details` in `social_core/pipeline/user.py`. While this might not seem relevant to some people, because everything works well if the server works correctly, I feel like security is at its best if both sides verify everything, even if it is "not that important".

My suggestion is to add this method to the `OpenIdConnect` class:

```
def process_refresh_token_response(self, response, *args, **kwargs):
    json = super().process_refresh_token_response(response, *args, **kwargs)
    id_token = json.get('id_token')
    if id_token is not None:
        self.validate_and_return_id_token(id_token, json['access_token'], check_nonce=False)
    return json
```

As a small note: Checking the nonce does not work anymore since it was deleted from the database earlier, so we would need to write a few lines of code to make sure it's not checked.

nijel:
Refresh responses could replace a trusted ID token without validating its signature or identity continuity. Validate refreshed tokens and bind their subject, audiences, authorized party, nonce, and authentication time before persistence.

Closes #589

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Security
+
+- OpenID Connect backends now validate ID tokens returned during token refresh
+  and reject changes to the authenticated identity.
+
 ## [5.0.2](https://github.com/python-social-auth/social-core/releases/tag/5.0.2) - 2026-06-26
 
 ### Security
--- a/social_core/backends/open_id_connect.py
+++ b/social_core/backends/open_id_connect.py
@@ -78,6 +78,7 @@ class OpenIdConnectAuth(BaseOAuth2PKCE):
     JWT_LEEWAY: float = 1.0  # seconds
     VALIDATE_AT_HASH: bool = True
     CUSTOM_AT_HASH_ALGO: str | None = None
+    ID_TOKEN_CONTEXT_KEY = "_oidc_id_token_context"
     # When these options are unspecified, server will choose via openid autoconfiguration
     ID_TOKEN_ISSUER = ""
     ACCESS_TOKEN_URL = ""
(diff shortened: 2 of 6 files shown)


davidhalter:
@nijel I'm not sure how well you have reviewed this, but I would be very skeptical that this actually does the correct things. 

LLMs seem to overfit these kind of scenarios and since I opened the original issue and have patched this in a code base to work, I'm not sure this is what we would want. I would personally revert this, but feel free to move forward with this change if you really think it's good. It just changes quite a few unrelated things, does add a lot of potentially unnecessary new code and just feels like slop. But I might be wrong. This is not to say that LLMs cannot be helpful in some cases...

cc @alakae You might want to know about this and maybe pin previous versions or review this, just to let you know.

nijel:
@davidhalter the original patch was pretty limited, but perhaps I shouldn't have followed all the review feedback. Will get back to this tomorrow...

davidhalter:
> the original patch was pretty limited

Why was it limited? I understand that it might not be optimal, but I think this OpenID Connect is a very security sensitive topic and I'm not sure I would want any change that is not clearly explainable.

nijel:
I started with simple ID token validation + checking that claims match the original token.

Additional code is from added claims validation. It probably should have been done in a separate pull request, but I folded it here because unvalidated original initial claims make validating refreshed claims harder (and this is part that was not covered in #589 at all).

Further additional complexity comes from `uid` not necessarily being the OIDC `sub` (Google OIDC and CAS derive it from other attributes) and from support to refresh pre-existing associations. This was indeed triggered by an LLM review, but I think these are valid concerns. We can certainly simplify the code by rejecting refreshes on pre-existing claims and removing all the migration logic. I will give it a try.

This also shows why maintaining this library is tricky. There are dozens of backends that do not always follow the same conventions, and most of these were one-off contributions years back with no ongoing maintenance from people that actually use them. LLM is great at pointing out such compatibility issues, but the decision on whether to try to be compatible at the cost of additional complexity is on me.

nijel:
https://github.com/python-social-auth/social-core/pull/1905 is the cleanup that removes some of the additional logic.

Issue #1176, opened 2025-05-26
Indentify and remove outdated backends


nijel:
This module has collected backends for years. Many of them are no longer working. I've randomly identified some of them when looking for backend documentation and removed them. But it would be great to go through all of them and remove the ones no longer available.

nijel:
itembase was removed in https://github.com/python-social-auth/social-core/pull/1411, the service is no longer available.

aviseth:
I ran a mechanical pass over all 158 backends that hardcode a URL and checked whether the hosts still resolve, then probed each `AUTHORIZATION_URL` over HTTP. Posting the results since "identify" is half of what this issue asks for.

Only DNS failures turned out to be worth anything. Most non-2xx responses are just a live endpoint rejecting a request with no `client_id` — `salesforce`, `stripe`, `twitch`, `strava` and friends all return 400 and are perfectly healthy — so I ignored status codes entirely.

**Dead: hardcoded host does not resolve, and no setting to override it**

- `skyrock` — `api.skyrock.com`
- `withings` — `developer.health.nokia.com` (and `health.nokia.com`)

Both are in #1939.

**Alive service, dead endpoint — wants a fix, not a removal**

- `monzo` — `auth.getmondo.co.uk` is gone (the pre-rename domain), but `api.monzo.com` is up.
- `mendeley` — `api-oauth2.mendeley.com` is gone, `api.mendeley.com` is up.

I left both alone rather than guess at replacement endpoints I cannot test against a real account. Happy to do them if you can confirm the current URLs, or if you would rather I just make the obvious change and let review catch it.

**Looks dead, is not**

- `osso` — `demo.ossoapp.com` does not resolve, but that is only the default for the `OSSO_BASE_URL` setting, so self-hosted deployments are unaffected.
- `etsy` — `api.etsy.com` is gone, but the backend already points at `openapi.etsy.com` for v3.
- `lastfm`, `openinfra`, `openstack` — the non-resolving strings are a docstring placeholder (`your.site`) and the `axschema.org` OpenID AX namespace, neither of which is an endpoint.

**What this method cannot see**

[...]

PR #1923, opened 2026-08-20
chore: drop defunct pocket, runkeeper, echosign, and exacttarget backends

aviseth:
Picking up #1176 (identify and remove outdated backends). These four backends all target services that no longer expose the OAuth endpoints they hardcode, so authentication against them can't succeed anymore:

- **pocket** — Mozilla shut Pocket down entirely in July 2025; the apps, web app, and `getpocket.com` API were removed and data purged. The `getpocket.com/v3/oauth` endpoints are gone. (No test file existed.)
- **runkeeper** — ASICS discontinued the RunKeeper Health Graph API for third-party integrations, which broke the `runkeeper.com/apps/authorize` OAuth flow this backend depends on. (No test file existed.)
- **echosign** — Adobe rebranded EchoSign to Adobe Sign back in 2015 and migrated off `echosign.com` to `adobesign.com`; the hardcoded `secure.echosign.com` authorize/token endpoints are retired. (No test file existed.)
- **exacttarget** — ExactTarget was acquired by Salesforce and became Salesforce Marketing Cloud; its legacy authentication (the JWT/pre-shared-key flow this backend implements) was retired on 2019-08-01. Removed its test alongside the backend.

Each backend is referenced only in its own module (and, for exacttarget, its own test) — `social_core/backends/__init__.py` is empty and there's no in-repo backend registry or docs to update, so removal is complete. The full test suite passes after the removals.

I've followed the pattern of the earlier removals (#1411 itembase, and the nk / rdio / shimmering drops) and left the `### Removed` CHANGELOG bullets for you to batch at release time, as with those — happy to add them here instead if you'd prefer.

I kept this to backends I'm confident are definitively dead. A few more are strong candidates but slightly more arguable (e.g. `vend`, now Lightspeed X-Series; `angel`, now Wellfound) — glad to send a follow-up if you'd like those pruned too.


nijel:
Can you please do matching edits in https://github.com/python-social-auth/social-docs/?

aviseth:
Done — python-social-auth/social-docs#484.

Only `pocket` and `runkeeper` had pages there; `echosign` and `exacttarget` were never documented. Nothing else in the docs tree referenced either page, so it's just the two files plus their `toctree` entries in `docs/backends/index.rst`.

nijel:

PR #1928, opened 2026-08-24
Stop sending access_token as a refresh_token

### LINKED ISSUE #826: OAuth2 grant type `refresh_token` sends the access token when no refresh token is available

aroh3006:
Fixes #826.

refresh_token() fell back to access_token when no refresh_token was stored, sending it to the token endpoint as if it were one. A refresh_token is optional per RFC 6749 4.1.4, and providers reject a request carrying the wrong kind of token. This always failed with a 401 instead of doing nothing.

Removed the fallback. This also exposed that Zoom and PayPal never actually saved a real refresh_token in the first place, since it wasn't listed in their EXTRA_DATA. Refresh never had a chance to work for those two even with a correct fallback removed. Added it there.

The AzureAD and AzureADB2C tests build their own mocked login response and it didn't include a refresh_token. Their test_refresh_token tests broke once the fallback was gone. Updated those fixtures to match what the providers actually return.

Added tests for the fixed fallback in test_storage.py. Ran the full test suite (1152 passed, 176 skipped, 8 xfailed) and ruff.

### DIFF
--- a/social_core/backends/paypal.py
+++ b/social_core/backends/paypal.py
@@ -19,6 +19,7 @@ class PayPalOAuth2(BaseOAuth2):
     DEFAULT_SCOPE = ["openid", "profile"]
     REFRESH_TOKEN_METHOD = "POST"
     REDIRECT_STATE = False
+    EXTRA_DATA = [("refresh_token", "refresh_token")]
 
     def user_data(self, access_token: str, *args, **kwargs) -> dict[str, Any] | None:
         auth_header = {"Authorization": f"Bearer {access_token}"}
--- a/social_core/backends/zoom.py
+++ b/social_core/backends/zoom.py
@@ -16,7 +16,10 @@ class ZoomOAuth2(BaseOAuth2):
     DEFAULT_SCOPE = ["user:read"]
     REFRESH_TOKEN_METHOD = "POST"
     REDIRECT_STATE = False
-    EXTRA_DATA = [("expires_in", "expires_in")]
+    EXTRA_DATA = [
+        ("expires_in", "expires_in"),
+        ("refresh_token", "refresh_token"),
+    ]
 
     def user_data(self, access_token: str, *args, **kwargs) -> dict[str, Any] | None:
         return self.get_json(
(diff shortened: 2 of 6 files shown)


nijel:
@codex review

PR #1939, opened 2026-08-28
chore: drop defunct skyrock and withings backends

aviseth:
Another batch for #1176, following the same shape as #1923.

Both backends hardcode their OAuth endpoints on hosts that no longer resolve, and neither exposes a setting to point them elsewhere, so neither can complete a login for anyone:

**skyrock** — every endpoint is on `api.skyrock.com`, which has no DNS record. `skyrock.com` and `www.skyrock.com` both still resolve, so the site is alive and it is specifically the developer API that is gone.

**withings** — all three OAuth1 endpoints are on `developer.health.nokia.com`, dating from when Nokia owned Withings. Neither that host nor `health.nokia.com` resolves. Withings bought itself back from Nokia in 2018 and now authenticates at `account.withings.com` over OAuth2, so this is not a URL swap: the OAuth1 flow `WithingsOAuth` implements is not offered any more. Repointing would mean rewriting it as an OAuth2 backend, which felt like a separate decision for you rather than something to slip into a removal.

Removes the two modules and `test_skyrock.py`. `withings` had no test. Nothing else in the tree referenced either name.

I checked the rest of the watchlist the same way and deliberately did **not** include several that look dead at first glance but are not:

- `osso` — the demo host is dead but `OSSO_BASE_URL` is a setting, so self-hosted installs still work.
- `monzo`, `mendeley` — their authorization hosts (`auth.getmondo.co.uk`, `api-oauth2.mendeley.com`) are dead, but both companies and APIs are alive at `api.monzo.com` and `api.mendeley.com`. These want fixing, not removing, and I did not want to guess at the replacement endpoints without being able to test a real flow.
- `etsy` — `api.etsy.com` is gone but the backend already uses `openapi.etsy.com` for v3.

Happy to open the monzo/mendeley repoint as a separate PR if you would like it, and to keep working through the list in batches this size.

`ruff check` and `ruff format --check` are clean. The suite is 1178 passed / 185 skipped, with 7 failures in `test_google.py::GoogleOneTapTest` that are identically present on an unmodified `master`.

I did not add a `CHANGELOG.md` entry, matching #1923 — there is no unreleased section open at the moment. Say the word and I will add one.


nijel:
Can you please also update https://github.com/python-social-auth/social-docs?

aviseth:
Done — python-social-auth/social-docs#499.

Removes both backend pages and their toctree entries, plus the Skyrock line and its link target in `intro.rst` (withings was never listed there). Same shape as the pocket/runkeeper docs removal that followed #1923. Happy to hold it until this one lands.

nijel:

PR #1940, opened 2026-08-29
Fix TypeError (500) on non-ASCII Discourse SSO signature

pacocartones:
### Problem

`DiscourseAuth.auth_complete` compares the SSO signature with:

```python
if not hmac.compare_digest(str(sso_signature), str(param_signature)):
```

`sso_signature` is the attacker-supplied `sig` request parameter. `hmac.compare_digest` raises `TypeError: comparing strings with non-ASCII characters is not supported` when a `str` argument contains non-ASCII, so a request with a non-ASCII `sig` triggers an unhandled exception (HTTP 500) instead of a clean auth failure.

### Fix

Use `constant_time_compare` from `social_core.utils` — the helper already used elsewhere in this project. It `.encode("utf-8")`s both sides to bytes before `hmac.compare_digest`, so it compares in constant time and returns `False` for a bad/garbage signature instead of raising.

One-line behavioural change; a non-ASCII `sig` now fails the check cleanly.


nijel:
@codex review

PR #1986, opened 2026-09-16
feat(backends): implement a generic CAS backend

sinavir:
This commit introduces support for the CAS protocol using python-cas library.

It is tested using zulip (for the client) and django-cas-server (for the cas server).

I'm not very confident about how to write the test for this. I'm interested to know if you recommend any code or documentation interesting to get that task done


nijel:
Thanks for the pull requests. The tests typically mock the authentication API provider, check existing tests for inspiration. Documentation should be added to https://github.com/python-social-auth/social-docs/.

nijel:
@codex review

pyproject.toml
```diff
@@ -48,6 +48,7 @@ dependencies = [
   "defusedxml>=0.7.1",
   "oauthlib>=3.3.1",
   "PyJWT[crypto]>=2.13.0",
+  "python-cas>=1.7.2",
```
nijel: This should probably go to an extra, not be part of core deps like Google OneTap or Shopify.

nijel:
Thanks for the adjustment, but this still lacks tests and documentation.

sinavir:
Yes, I'm currently very busy. It will come at some point.

